PR #29, opened 2014-01-16
added the ability to set a host's host to false

### LINKED ISSUE #28: Allow Relative Path Urls
We currently use grunt with grunt-connect-proxy for local development so that we can code our front end to use relative path AJAX requests to avoid cross site request issues.

When we deploy the application to other environments we simply configure the load balancer to perform the same type of proxying of relative path requests to other servers/ports server side.

I'm attempting to configure the esFactory as such:

```
angular.module('components.search')
  .service('es', function (esFactory) {
    return esFactory({
      host: '/services/elastic'
    });
  });
```

and using elasticsearch.angular.js but the AJAX requests are being made to http://localhost/services/elastic instead of /services/elastic

This is causing collisions with the way grunt runs the application in development on a url like http://localhost:9000/#/

Can we update the code to not assume I want to prefix the AJAX requests with http://localhost if I specify a host with a a leading slash?

This might just be achievable through a new configuration param:

```
angular.module('components.search')
  .service('es', function (esFactory) {
    return esFactory({
      host: '/services/elastic',
      relative: true
    });
  });
```

and specifically handled in elasticsearch.angular.js when making the call with $http

```
 this.$http({
    ...
    url: this.host.makeUrl(params),
    ...
  })
```

spalger:
Allows the use of relative urls, closes #28

### DIFF
--- a/src/lib/host.js
+++ b/src/lib/host.js
@@ -135,7 +135,11 @@ Host.prototype.makeUrl = function (params) {
     auth = this.auth + '@';
   }
 
-  return this.protocol + '://' + auth + this.host + port + path + (query ? '?' + query : '');
+  if (this.host) {
+    return this.protocol + '://' + auth + this.host + port + path + (query ? '?' + query : '');
+  } else {
+    return path + (query ? '?' + query : '');
+  }
 };
 
 Host.prototype.toString = function () {
--- a/test/unit/test_host.js
+++ b/test/unit/test_host.js
@@ -149,6 +149,16 @@ describe('Host class', function () {
       host = new Host({ host: 'italy', path: '/pie', auth: 'user:pass'});
       expect(host.makeUrl()).to.be('http://user:pass@italy:9200/pie');
     });
+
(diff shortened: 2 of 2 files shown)


andreimatusevich:
It doesn't work in elasticsearch.angular.js because of these lines 
```javascript
    if (!config.hosts && !config.host) {
      config.host = 'http://localhost:9200';
    }
```

spalger:
@andreimatusevich the purpose of this change was to allow this:
```js
var client = new Client({
  hosts: [
    {
      host: false,
      path: '/elasticsearch-endpoint',
      port: 443
    }
  ]
})
```

If neither host or port are defined the default is supposed to be `http://localhost:9200`

andreimatusevich:
I thought this feature was about to send request to the current host whatever it is
```javascript
if (this.host) {
    return this.protocol + '://' + auth + this.host + port + path + (query ? '?' + query : '');
} else {
    return path + (query ? '?' + query : '');
}
```
It would be nice to support relative urls

PR #150, opened 2014-10-08
Fix non-working example in quick_start

tanguylebarzic:
- Wrap the 'filtered' object in a top-level query object (otherwise same issue as https://github.com/elasticsearch/elasticsearch/issues/1688)
- Fix typo (err => error) in the callback.
It's nice to be able to start from a working example and iterate from here. Thanks!


spalger:
Closing due to lack of CLA signature, please let me know if you have in fact signed it please let me know.

PR #176, opened 2015-01-08
Fixed opening bracket for parameter - client.count

quilesbaker:
The opening bracket was missing for the json object passed as first parameter. On the api docs. See diff.


spalger:
Closing due to lack of CLA signature, please let me know if you have in fact signed it please let me know.

PR #191, opened 2015-02-24
Mistake in count + filtered exemple

samiadvize:
When using filtered it needs to be in a query not body or you get an error message like 

```
'BroadcastShardOperationFailedException[[indexName][1] ]; nested: QueryParsingException[[indexName] request does not support [filtered]];
```

+ an opening { was missing

### DIFF
--- a/docs/api_methods.asciidoc
+++ b/docs/api_methods.asciidoc
@@ -233,9 +233,9 @@ client.count({
 .Get the number of documents matching a query
 [source,js]
 ---------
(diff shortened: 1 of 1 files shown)


spalger:
Sorry I never provided feedback here, but the commit never updated when you signed the CLA and values put at the top level of the params are placed in the query string (not the query part of the body). In this case it probably sent `?query=[object Object]` to elasticsearch.

PR #192, opened 2015-02-25
Update quick_start.asciidoc

benx4:
If there is no protocol here, you may get a connection error here


spalger:
Closing due to lack of CLA signature

PR #206, opened 2015-04-08
Fix/205

### LINKED ISSUE #205: Browser clients do not use host-specified auth
Currently, if you specify basic auth information in the host object it is not used by some of the browser connectors (verified in xhr connector).

**Workaround:** use the `Authorization` header:
```js
var client = new elasticsearch.Client({
  host: {
    host: 'localhost',
    port: 9200,
    headers: {
      'Authorization': 'Basic ' + btoa(user + ':' + password)
    }
  }
});
```

spalger:
Fixes #205 

This moved all authentication juggling out of the connectors and handles it generically using headers in the `Host` class.

 - flatten auth config to an Authorization header when the Host is created
 - remove individual Authorization handling from connectors
 - removed incomplete support for per-request auth
 - use per-request headers to provide your own Authorization header per request

### DIFF
--- a/scripts/generate/index.js
+++ b/scripts/generate/index.js
@@ -31,7 +31,7 @@ var path = require('path');
 var fromRoot = path.join.bind(path, require('find-root')(__dirname));
 var utils = require(fromRoot('grunt/utils'));
 var _ = require(fromRoot('src/lib/utils'));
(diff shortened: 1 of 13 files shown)


lukasolson:
LDTM, where D = Delicious :yum:

spalger:
🎂 YUM! 🎂

PR #223, opened 2015-06-08
Update quick_start.asciidoc

clintongormley:
Fixed the example with the filtered query, missing the top-level `query` parameter

See https://github.com/elastic/elasticsearch/issues/1688#issuecomment-109646788


spalger:
Awesome, thanks @clintongormley

PR #235, opened 2015-07-01
Adding missing comma

ycombinator:


spalger:
Thanks @ycombinator, can you apply these changes to the [examples source](https://github.com/elastic/elasticsearch-js/blob/master/docs/_examples/msearch.asciidoc) and regenerate the docs please?

ycombinator:
@spalger I've never messed with the docs before. Should I *only* apply this change to the example doc or is that in *addition* to the change I've already made?

spalger:
If you apply it to the example file, and then run `npm run generate` it will rebuild the file that you changed here with the updated example (as well as the other doc pages using that example)

ycombinator:
@spalger Thanks. I did as you said but I'm afraid the generated artifacts include things unrelated to my simple typo fix. Is this expected?

spalger:
Well, yes and no. The generate command updates the API as well as the docs, but it looks like some of the automatic updates don't regenerate nicely (specifically the snapshot.create method).

If you wouldn't mind just pushing the updated doc files that would be appreciated.

ycombinator:
@spalger Done; this looks much better. Would you like me to squash my commits for a cleaner history?

spalger:
No need :) I like an accurate history!

PR #237, opened 2015-07-14
Update elasticsearch-js msearch body format

cyanCYMK:
Although _msearch using the web API will allow the body to not have leading underscores before the keys "index" and "type" (e.g. `{"index": "my-index", "type": "my-type"}
{"query": ... }`), the Javascript Elasticsearch client.msearch will give the error "SearchPhaseExecutionException[Failed to execute phase [query], all shards failed]" if the body follows the same format. Adding the leading underscores fixes this issue.


cyanCYMK:
I have signed the CLA, maybe I submitted the pull request too soon after signing.

PR #248, opened 2015-08-04
Added support for http error 429: 'Too Many Requests'

Zenedith:
Added support for http error 429: 'Too Many Requests', eg. for bonsai.io


spalger:
Gratzi!

PR #249, opened 2015-08-05
updated client.count example

zoellner:
current example leads to error 400, QueryParsingException, request does not support [filtered]]


spalger:
Ah, good catch. thanks!

PR #252, opened 2015-08-06
correct typo

msimerson:


msimerson:
CLA is signed. You might need to refresh something for that to show up.

spalger:
I see it on my side. Thanks!

PR #266, opened 2015-09-10
Get rid of the number of the beast

dropfen:
This commit just changes the numbers 666, which is used on some places in the tests and docs to the number 777


spalger:
I'd be happy to merge this if you don't mind [signing the CLA](http://www.elasticsearch.org/contributor-agreement/) :) Thanks!

dropfen:
ok men, it's done ;)

PR #275, opened 2015-10-06
Fixing domain callbacks

lpetre:
We're in the process of converting from postgres to elasticsearch for search.

This module's connection pool is breaking our use of node's domains. Because connections are created with one domain and might service a request for a different domain later when they're reused.

I've modeled my fix after similar fixes in node-pg:
https://github.com/brianc/node-postgres/commit/779c8064f2873b190d6f671b6ecef75852688526
https://github.com/brianc/node-postgres/commit/9c87253aff3609dd3444d53ce5878b1112c7c646

Please let me know what you think.

### DIFF
--- a/src/lib/transport.js
+++ b/src/lib/transport.js
@@ -140,6 +140,10 @@ Transport.prototype.request = function (params, cb) {
 
   // determine the response based on the presense of a callback
   if (typeof cb === 'function') {
+    // handle callbacks within a domain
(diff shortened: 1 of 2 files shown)


lpetre:
FWIW I have signed the CLA.

spalger:
Weird, I tracked down your signature but it didn't make it into our tooling. Sorry about that.

PR #276, opened 2015-10-07
updated the browser dependancy for lodash

DrHofman:
When using the module with browserify it fails to load the lodash dependancy with the latest version of npm, removing the relative path resolves this issue


spalger:
:+1: thanks

PR #303, opened 2015-11-17
Update quick start guide's complex query.

Friss:
Allows the code to be used without errors.

### DIFF
--- a/docs/quick_start.asciidoc
+++ b/docs/quick_start.asciidoc
@@ -113,49 +113,50 @@ NOTE: In this example, `request` and `response` are http://expressjs.com/api.htm
 
 [source,js]
 -----------------
-var pageNum = request.param('page', 1);
-var perPage = request.param('per_page', 15);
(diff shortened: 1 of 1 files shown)


spalger:
I'm not opposed to using the `request.params` syntax vs the deprecated `request.param()` helper, but I don't understand the rest of the changes. It also looks like there is a [syntax error](https://github.com/Friss/elasticsearch-js/blob/update-quick-start/docs/quick_start.asciidoc#more-complex-filtered-query) in one of the params.

spalger:
Also, I see that you have signed the CLA. thank you

Friss:
@spalger Ah missed a `'` when removing the old params. 

The other changes are 

* Adding in a missing `}` before the callback function
* Renaming the ES response variables to avoid a potential clash with Express's variables. (`request`, `response` for Express and `err` and `res` for ES).

spalger:
Thanks :)

Issue #260, opened 2015-09-04
When logging a host, credentials should not be included


spalger:
Currently, when logging "Unable to revive connection: http://user:password@localhost:9200/" the plaintext username and password is included.

delvedor:
Fixed in the new client 🎉

PR #304, opened 2015-11-17
user:password@ removal on log warning

manonthemat:
based on the described issue https://github.com/elastic/elasticsearch-js/issues/260, this pull request has a simple solution of logging a connection id without the user:password combination.

`Unable to revive connection: http://user:password@localhost:9200/` will turn into `Unable to revive connection: http://localhost:9200/`

### DIFF
--- a/src/lib/connection_pool.js
+++ b/src/lib/connection_pool.js
@@ -222,7 +222,7 @@ ConnectionPool.prototype._selectDeadConnection = function (cb) {
     if (timeout.conn.status === 'dead') {
       timeout.revive(function (err) {
         if (err) {
-          log.warning('Unable to revive connection: ' + timeout.conn.id);
+          log.warning('Unable to revive connection: ' + timeout.conn.id.replace(/\w+:\w+@/, ''));
           process.nextTick(next);
         } else {
           cb(void 0, timeout.conn);


manonthemat:
the used regular expression captures username and passwords that are composed of letter, number, underscore. According to the W3 spec, a username or password can have a length of 0 and consist of "!", "$", "&", "'", "(", ")", "*", "+", ",", "-", ".", "/", ":", ";", "=", "?", "@", "_", "~", and code points in the ranges U+00A0 to U+D7FF, U+E000 to U+FDCF, U+FDF0 to U+FFFD, U+10000 to U+1FFFD, U+20000 to U+2FFFD, U+30000 to U+3FFFD, U+40000 to U+4FFFD, U+50000 to U+5FFFD, U+60000 to U+6FFFD, U+70000 to U+7FFFD, U+80000 to U+8FFFD, U+90000 to U+9FFFD, U+A0000 to U+AFFFD, U+B0000 to U+BFFFD, U+C0000 to U+CFFFD, U+D0000 to U+DFFFD, U+E0000 to U+EFFFD, U+F0000 to U+FFFFD, U+100000 to U+10FFFD.

Therefore the regex needs adjustment or a different solution should be used altogether.

spalger:
Yeah, I would rather change the value that we use (currently `connection.id`). Perhaps we can just add a `description` to the connection which can be used to describe the connection in scenarios like this.

PR #328, opened 2016-01-04
allow to set http.agent to false by setting noAgent to true in client…

### LINKED ISSUE #327: set http.agent to false
Due to [this issue](https://github.com/nodejs/node-v0.x-archive/issues/6833) it is necessary to have the possibility to set the http.agent to `false`. While doing a bulk insert into elastic I get `no living connections available` aka `EADDRNOTAVAIL`.

jochenonline:
This adds the new option `noAgent` which, if set to true, sets the http.agent to `false`. This fixes #327.

### DIFF
--- a/src/lib/connectors/http.js
+++ b/src/lib/connectors/http.js
@@ -44,7 +44,7 @@ function HttpConnector(host, config) {
     maxSockets: 11
   });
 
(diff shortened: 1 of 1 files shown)


spalger:
Closed in favor of #329

Issue #226, opened 2015-06-13
Showing hostname in log messages?


julien51:
We're trying to debug seemingly "random" errors that we get inside our ElasticSearch Client and I have to say that The JS library makes our job quite difficult:

```javascript
Elasticsearch ERROR: 2015-06-13T14:16:53Z
  Error: Request error, retrying -- socket hang up
      at Log.error (/home/node/trends/node_modules/elasticsearch/src/lib/log.js:213:60)
      at checkRespForFailure (/home/node/trends/node_modules/elasticsearch/src/lib/transport.js:192:18)
      at HttpConnector.<anonymous> (/home/node/trends/node_modules/elasticsearch/src/lib/connectors/http.js:153:7)
      at ClientRequest.wrapper (/home/node/trends/node_modules/elasticsearch/node_modules/lodash/index.js:3128:19)
      at ClientRequest.emit (events.js:95:17)
      at CleartextStream.socketCloseListener (http.js:1522:9)
      at CleartextStream.emit (events.js:117:20)
      at tls.js:691:10
      at process._tickCallback (node.js:419:13)
```

`Request error, retrying -- socket hang up` is quite a limited error message :(

It would be really convenient to have more details :) Things like the hostname (when talking to a cluster, that may help look into the log of the right server...), but also, ideally the last request?... etc

spalger:
I agree, we should do more here

saurabhjaluka:
Any update on this issue? as I am facing the similar problem.

spalger:
Nothing to report now. Will try to take a look soon but this week is pretty packed.

If either of you are interested in giving or a shot the errors are located in [src/lib/errors.js](https://github.com/elastic/elasticsearch-js/blob/master/src/lib/errors.js).

saurabhjaluka:
Okay, thanks. I will try to look into it.

jmm:
@spalger I took a crack at this in #330. `src/lib/errors.js` didn't appear to play a role in this scenario from what I could tell.

spalger:
Closed by #330

sahas-:
Sorry, to comment on a closed issue. I'm seeing the same error with the latest stable release & also on v13.0.0-beta2.. can we re-open this please?

Elasticsearch ERROR: 2017-03-08T03:24:46Z
  Error: Request error, retrying
  POST http://xxxxx/_search => socket hang up
[...]

PR #330, opened 2016-01-06
Enhance HTTP connector / transport error messages with more detail

jmm:
See #226. Another approach would be to construct the request information in `src/lib/transport` and leave `src/lib/connectors/http` untouched, if it's preferable for some reason. I don't have a great understanding of how the connectors / transport interact. Either way, [this](https://github.com/jmm/elasticsearch-js/blob/7bc7d0becbd42255d7d5c0187664f217dbdf28be/src/lib/connectors/http.js#L154-L163) is duplicating code from [`src/lib/log`](https://github.com/elastic/elasticsearch-js/blob/0e3fe1586fd867275fd0e9abb7126724be3d3761/src/lib/log.js#L292-L298) for massaging the request data to be suitable input for `url.format()`. So perhaps that should be extracted and shared.

I'm checking this with something like:

```js
new elasticsearch.Transport({
  host: 'whatever.example:9200',
  log: 'error',
})
.request({
  method: "GET",
  path: "/whatever",
})
```

(That's not the greatest example because the existing error message for that actually includes the hostname, but just imagine it's `... => socket hang up`.)

### DIFF
--- a/src/lib/transport.js
+++ b/src/lib/transport.js
@@ -226,12 +226,24 @@ Transport.prototype.request = function (params, cb) {
 
     if (err) {
       connection.setStatus('dead');
+
+      var errMsg = err.message || '';
+
+      errMsg =
+        "\n" +
+        params.req.method +
+        ' ' +
+        connection.host.makeUrl(params.req) +
+        (errMsg.length ? ' => ' : '') +
+        errMsg
+      ;
+
       if (remainingRetries) {
         remainingRetries--;
-        self.log.error('Request error, retrying' + (err.message ? ' -- ' + err.message : ''));
+        self.log.error('Request error, retrying' + errMsg);
         self.connectionPool.select(sendReqWithConnection);
       } else {
-        self.log.error('Request complete with error' + (err.message ? ' -- ' + err.message : ''));
+        self.log.error('Request complete with error' + errMsg);
         respond(new errors.ConnectionFault(err));
       }
     } else {


spalger:
Hmm, I agree that we should probably move [this](https://github.com/elastic/elasticsearch-js/blob/0e3fe1586fd867275fd0e9abb7126724be3d3761/src/lib/log.js#L292-L298) into a helper of some sort.

As for the transport/connector bit, anything that should happen in all scenarios should happen in the transport. It is always used, and the connectors are swapped out in certain scenarios. In this implementation the error messages will only be enhanced for node.js and browsers will not see the hostname.

jmm:
> Hmm, I agree that we should probably move this into a helper of some sort.

This turned out to be complicated by nuances and inconsistencies of how strings and objects representing URLs are interpreted and structured by various modules interacting here, but ultimately I wound up with a simpler solution: `host.makeUrl()` (see new commits).

> As for the transport/connector bit[...]

Ok, that makes sense. I originally had it that way and moved it into http because the request data was more readily available there and I wasn't sure it'd make a difference. For example, I didn't want to have to duplicate [this](https://github.com/elastic/elasticsearch-js/blob/7d5b0395e65a4971ce1df5dd618d6d8376f69914/src/lib/connectors/http.js#L111) kind of logic. But it seems that `Host.prototype.makeUrl()` is suitable for this use case with the request data that's already accessible in transport.

Is splitting the output to multiple lines like this ok here?

I'm checking the output with something like:

```js
new elasticsearch.Transport({
  host: {
    host: "whatever.example",
    port: 9200,
    path: "/host-path",
    query: "host=h",
  },
  log: 'trace',
})
.request({
  method: "GET",
  path: "/request-path",
  query: "request=r",
})
```

spalger:
This is great @jmm 

for posterity: https://gist.github.com/spalger/6cb2c1c41fb8dceeb702

jmm:
Thanks!

PR #334, opened 2016-01-08
Fix Transport request doc issues.

jmm:
`url` does not seem to be a valid param at this time.

In the asciidoc `query` is only documented as accepting an object. In practice, passing a string seems to be supported. Is that something you wish to document, or should I edit that out of `transport.js`?


spalger:
:heart:

PR #336, opened 2016-01-08
Perform early return in Log ctor.

jmm:
Just a style suggestion, that makes sense to me here.

[Nicer diff](https://github.com/elastic/elasticsearch-js/pull/336/files?w=1).


spalger:
:+1: Thanks

PR #346, opened 2016-01-20
Update host.js => window.location is not defined

damienleroux:
window can be defined and yet window.location can be undefined: both "window" and  "window.location" must be defined to access to "protocol" information.

I encountered this problem where "window" was defined but not "window.location". This append when transpiling js file in a express-react-browserify env.

Regards

### DIFF
--- a/src/lib/host.js
+++ b/src/lib/host.js
@@ -12,7 +12,7 @@ var startsWithProtocolRE = /^([a-z]+:)?\/\//;
 var defaultProto = 'http:';
 var btoa;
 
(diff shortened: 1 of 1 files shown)


spalger:
@damienleroux thanks for the patch!

Please consider signing the CLA so that we can review this pull

damienleroux:
It is done; I signed it.

damienleroux:
Hi,

Could you confirm the CLA worked and complete the merge please?

Regards

spalger:
LGTM

Issue #137, opened 2014-09-01
Re-configure client host dynamically


nocturnal3d:
Hi,

I'm using AngularJS with elasticsearch-js and I can't find a way of setting the esFactory/Client host dynamically.

I want to provide a user setting to configure the Client hostname dynamically. Is there a way I can re-configure the service when the user types in a different hostname for ES?

thanks in advance.

maneesh54321:
I am getting:  TypeError: Invalid host
    at AngularConnector.ConnectionAbstract (elasticsearch.angular.js:12188)
    at new AngularConnector (elasticsearch.angular.js:291)
    at ConnectionPool.setHosts (elasticsearch.angular.js:16656)
    at ChildScope.$scope.checkElastic (kpiElastic.js:133)
    at fn (eval at compile (angular.js:13365), <anonymous>:4:227)
    at callback (angular.js:23613)
    at ChildScope.$eval (angular.js:16052)
    at ChildScope.$apply (angular.js:16152)
    at HTMLButtonElement.<anonymous> (angular.js:23618)
    at HTMLButtonElement.dispatch (jquery.min.js:3)

Syntax used:   es.transport.connectionPool.setHosts([
				   $scope.url  //this value has to be passed through UI
				]);
//es is the factory instance injected in controller.

PR #352, opened 2016-01-31
Fix  #324

### LINKED ISSUE #324: Re-configure client host dynamically
I've found #137 and tried to use such approach in my app, but have reached another problem:
I'm using elasticsearch - v10.0.1 - 2015-11-30 with elasticsearch.angular.js and method, described in the #137  run me into Type error, due to this check in elasticsearch.angular.js (part of function `ConnectionAbstract(host, config)` function):
```javascript
  if (!host) {
    throw new TypeError('Missing host');
  } else if (host instanceof Host) {
    this.host = host;
  } else {
    throw new TypeError('Invalid host');
  }
```
And, actually, if I'm passing new host as string, like described above, I'll not pass `host instanceof Host` check, but, I also cannot create instance of Host object (I simply cannot access that type).

I'm trying to do smth like this:
```javascript
elasticsearchClient.$inject = ['esFactory', 'configurationResource'];
  function elasticsearchClient(esFactory, configurationResource) {
    var host = 'http://old.host.com:6901/';
    var apiVersion = '1.7';
    var client = esFactory({
      hosts: [host],
      apiVersion: apiVersion
    });
    configurationResource.configValues.get({
      fullName: 'ElasticsearchUrl'
    }).$promise.then(function(result) {
      var newHost = result.currentValue; // newHost==='http://new.host.com:6901/'
      client.transport.connectionPool.setHosts([newHost]);
    });

    return client;
  }
```
Maybe I'm doing smth wrong ? Could you please help me ?

xSAVIKx:
Added support for string hosts in ConnectionAbstract constructor.
Host constructor is used with supplied host string to create host object for ConnectionAbstract.
With this fix it will be possible again to configure ES connection dynamically as was described in #137

### DIFF
--- a/src/lib/connection.js
+++ b/src/lib/connection.js
@@ -22,6 +22,8 @@ function ConnectionAbstract(host, config) {
     throw new TypeError('Missing host');
   } else if (host instanceof Host) {
     this.host = host;
+  } else if (typeof host === "string"){
+    this.host = new Host(host);
   } else {
     throw new TypeError('Invalid host');
   }
--- a/test/unit/specs/connection_abstract.js
+++ b/test/unit/specs/connection_abstract.js
@@ -15,6 +15,15 @@ describe('Connection Abstract', function () {
     expect(conn.host).to.be(host);
   });
 
+  it('constructs with string host using Host constructor', function(){
+    var host = 'localhost';
+    var port = 9200;
+    var hostUrl = host+':'+port;
+    var conn = new ConnectionAbstract(hostUrl);
+    expect(conn.host.host).to.be(host);
+    expect(conn.host.port).to.be(port);
+  });
+
   it('requires a valid host', function () {
     expect(function () {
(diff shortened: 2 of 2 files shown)


xSAVIKx:
Strange, but I've signed CLA...

spalger:
Thanks @xSAVIKx! I merged your changes but updated them to run using the Transport instead. This way the hosts created will reflect the options specified on client initialization (for things like [suggest compression](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/configuration.html#config-suggest-compression) for instance).

spalger:
I should be able to release these later today

xSAVIKx:
Great :-)
So, if I understand your changes, now we'll be able to do following:
```
client.transport.setHosts([
  'http://someOtherHost:9200'
]);
```
If so - Thanks a lot :+1:

spalger:
yep!

PR #361, opened 2016-02-13
Fix dependence promise-js conflict with core-js

piaolingxue:
lazy load promise-js 

then use other promise library replace it

```js
import Bluebird from 'bluebird';
export const client = new elasticsearch.Client({
  host: 'localhost:9200',
  log: 'warning',
  defer: function () {
    var resolve, reject;
    // Bluebird.defer() is deprecated @ http://bluebirdjs.com/docs/api/deferred-migration.html
    var promise = new Bluebird(function() {
      resolve = arguments[0];
      reject = arguments[1];
    });
    return {
      resolve: resolve,
      reject: reject,
      promise: promise
    };
  }
});
```

### DIFF
--- a/src/lib/transport.js
+++ b/src/lib/transport.js
@@ -7,7 +7,6 @@ module.exports = Transport;
 var _ = require('./utils');
 var errors = require('./errors');
 var Host = require('./host');
-var Promise = require('promise-js');
 var patchSniffOnConnectionFault = require('./transport/sniff_on_connection_fault');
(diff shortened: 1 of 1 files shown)


piaolingxue:
image
Sorry, I can't find the reason why sometimes build success but sometimes build failure.

piaolingxue:
jenkins, test it

spalger:
jenkins, test it

spalger:
Went with #362 instead. Thanks!

Issue #345, opened 2016-01-20
dependence promise-js conflict with core-js


jseagull:
```sh
> require('promise-js')
{ [Function: Promise]
  all: [Function],
  cast: [Function],
  race: [Function],
  reject: [Function],
  resolve: [Function] }
> require('core-js')
TypeError: Cannot redefine property: Symbol(Symbol.iterator)
    at Object.defineProperty [as setDesc] (native)
    at module.exports.object.(anonymous function) (/home/seagull/huaban/muse/node_modules/core-js/modules/$.hide.js:4:12)
    at module.exports (/home/seagull/huaban/muse/node_modules/core-js/modules/$.iter-define.js:50:5)
    at Object.<anonymous> (/home/seagull/huaban/muse/node_modules/core-js/modules/es6.array.iterator.js:11:44)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
```

jseagull:
https://github.com/zloirock/core-js/issues/161#issuecomment-173079010

ilionic:
Also getting this error after pulling latest version `elasticsearch` (>10), conflicting with other libraries.
Rolling back to `9.0.2`
```
anonymous/node_modules/react-helmet/node_modules/core-js/modules/_hide.js:4
  return $.setDesc(object, key, createDesc(1, value));
           ^

TypeError: Cannot redefine property: Symbol(Symbol.iterator)
    at Object.defineProperty [as setDesc] (native)
    at module.exports.object.(anonymous function) (anonymous/node_modules/react-helmet/node_modules/core-js/modules/_hide.js:4:12)
    at module.exports (anonymous/node_modules/react-helmet/node_modules/core-js/modules/_iter-define.js:54:5)
    at Object.<anonymous> (anonymous/node_modules/react-helmet/node_modules/core-js/modules/es6.array.iterator.js:11:43)
    at Module._compile (module.js:398:26)
    at Module._extensions..js (module.js:405:10)
    at Object.require.extensions.(anonymous function) [as .js] (anonymous/node_modules/babel-register/lib/node.js:138:7)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
[...]

PR #362, opened 2016-02-15
[promise] swap out promise-js

spalger:
due to https://github.com/elastic/elasticsearch-js/issues/345 we can't use promise-js. This replaces it with npm's `promise/lib/es6-extensions` which provides a bare-bones implementation with just the es6 extensions and doesn't pollute global anything.


satazor:
This would be awesome.. I've just hit another issue with `promise-js`, see. https://github.com/kevincennis/promise/issues/2

spalger:
jenkins, test it

spalger:
jenkins, test it

spalger:
jenkins, test it

Issue #393, opened 2016-04-21
Support retrying after a requestTimeout


mhahn:
I'm running into requests periodically timing out. I'd like to set a low timeout value and then have the request get retried up to maxRetries if it times out. Right now if a timeout is hit, the request aborts.

mhahn:
I guess `requestTimeout` is for the entire request -- even if there are failures and retries. Maybe what I'm looking for is out of the scope of what this library wants to provide.

Essentially I'd like to set some very small timeout (like 100ms) and if the request doesn't return in that time, try again up to maxRetries.

Any ideas on supporting something like that?

mhahn:
i'm going to close this unless someone else thinks its useful. i'd actually prefer to just set a smaller requestTimeout and handle this outside the library.

PR #394, opened 2016-04-22
Add support for a `connectionTimeout`

mhahn:
This allows us to specify a timeout for how long we should wait for ES
to execute a request. If the request takes longer than
`connectionTimeout`, the request will be retried up to `maxRetries`.

An attempt at: https://github.com/elastic/elasticsearch-js/issues/393


mhahn:
i'm not a huge fan of this approach. i like the simplicity of `requestTimeout` i'll use that to add something like this outside the library.

PR #399, opened 2016-05-03
Switch to webpack

spalger:
I'm not sure what happened, but at some point the browser builds stopped minimizing correctly. I'm trying webpack out as I have a lot more experience with it and think it might be easier for me to maintain.


spalger:
jenkins, test it

PR #411, opened 2016-06-01
fixes typo in quick start doc

Cspeisman:
there was a missing t for the word against on line 130 of the quick start document

line use to read:
// match the query agains all of


spalger:
Thanks!

PR #428, opened 2016-07-20
Pass the Client instance to ClientAction namespaceFactory

w33ble:
This PR makes the client instance available to namespaced plugins.

I was planning to add tests, but it doesn't look like there are currently any tests for the namespace stuff.


spalger:
LGTM

PR #448, opened 2016-09-22
Add port numbers and protocols to code examples

ctindel:


spalger:
Sorry, for letting this sit here, but I don't think that 9243 is a standard elasticsearch port, is it?

ctindel:
@spalger Well the docs say it can allocate any port between 9200 and 9300.  Port 9243 is the number allocated to my instance on cloud.elastic.com for use with https.

I think if you want to use a different number for the docs that's OK with me I just think we should show something other than 9200 since that's kind of a default for insecure http connections.

spalger:
> I just think we should show something other than 9200 since that's kind of a default for insecure http connections

That may be the case in your setup, but I'm pretty sure that when you turn on https in elasticsearch it still uses port 9200, unless specifically configured otherwise. Because of this I'm not comfortable changing the port in the docs

ctindel:
Well, maybe don't take the first line in the change, but the following examples doesn't show using https or a port numbers.  Maybe you can take what I did but just change everything to 9200 so people looking for an example of how to use something besides 9200, which will basically include everybody using elastic cloud, have something to look at.  As it was I had to ask an engineer here how to do it because it wasn't documented anywhere.

PR #482, opened 2016-12-20
Allows plugins to be added after client instantiation

tylersmalley:


tylersmalley:
Closing for now as it doesn't currently work with `namespaceFactory`

PR #488, opened 2017-01-05
Update api_methods.asciidoc

TonyApuzzo:
scrollId is the wrong parameter name for clearScroll, scroll_id works.


TonyApuzzo:
Put in a bug report since I don't have time to deal with the CLA.

PR #494, opened 2017-01-26
updated package.json to include keywords

Abykin:
I noticed that you were missing the keywords in your package.json file, so I've added them. Let me know if there are other/different keywords that you'd use instead.


spalger:
Thanks!

PR #523, opened 2017-04-04
export console logger in non browser environments

tellnes:
When using this library in eg. AWS Lambda environment the console logger
is a lot more usefull than stdio.


spalger:
Cool, thanks @tellnes

spalger:
Just pushed out `13.0.0-rc1`

PR #535, opened 2017-04-25
Manually incrementing the hardcoded master version to 6.0

kobelb:
The `master` branch was hard-coded to be version 5.0 in the Version.js, this just increments that number.

The other approach that I was considering was adding in an optional 'branch to version' in the package.json to hopefully make it more obvious next time. Open to other alternative as well.

### DIFF
--- a/scripts/Version.js
+++ b/scripts/Version.js
@@ -1,6 +1,7 @@
 var _ = require('lodash');
 var pkg = require('../package.json');
 var branches = pkg.config.supported_es_branches;
+var branchVersions = pkg.config.branch_versions;
 var semver = require('semver');
 
 var maxMinorVersion = function (majorV) {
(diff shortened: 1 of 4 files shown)


scripts/Version.js
```diff
@@ -19,7 +19,7 @@ Version.fromBranch = function (branch) {
   var m;
 
   // master === the highest version number
-  if (branch === 'master') return new Version('5.0.0');
+  if (branch === 'master') return new Version('6.0.0');
```
spalger: Ugh... I'm not totally opposed to putting this in package.json's config section

spalger:
I assume you also ran `npm run generate`?

kobelb:
@spalger yup, I ran `npm run generate` and it only changed 5.x and master.

scripts/Version.js
kobelb: Cool, lemme throw something together for this!

kobelb:
@spalger threw together a version that uses the package.json to drive this, lemme know what you think.

scripts/Version.js
```diff
@@ -17,16 +18,17 @@ function Version(v) {
...
+  var majorMinorRegex = /^\d+\.\d+$/;
 
   // n.m -> n.m.0
-  if (m = branch.match(/^\d+\.\d+$/)) return new Version(branch + '.0');
+  if (m = branch.match(majorMinorRegex)) return new Version(branch + '.0');
 
   // n.x -> n.(maxVersion + 1).0
   if (m = branch.match(/^(\d+)\.x$/i)) return maxMinorVersion(m[1]).increment('minor');
 
+  var branchVersion = branchVersions[branch];
```
kobelb: We might want to use this before the other cases.. on second thought, lemme change this...

spalger:
Thanks @kobelb

spalger:
BTW, I was debugging b345e0b13d0b00b5b37a861e4839eb31d6eb81eb and ended up defining the master branch translation just like we do the translation for `5.x`.

To get the next major version:
 - for the list of branches in `pkg.config.supported_es_branches` and `pkg.config.unstable_es_branches`
    - split branch name on `.`
    - take the first part
    - parse it to a number
 - filter out any `NaN`s (excludes "master" at this point)
 - sort versions descending
 - take the first version and add `1`

To get the next minor version within a major version:
 - for the list of branches in `pkg.config.supported_es_branches` and `pkg.config.unstable_es_branches`
    - split branch name on `.`
    - parse all parts to a number
    - take the first two parts
 - filter tuples:
    - where `tuple.length !== 2` (excludes "master")
    - where any values are `NaN` (excludes `.x` branches)
    - where `tuple[0]` is does not match the desired major version
 - sort tuples by `tuple[1]` descending
 - take `tuple[1]` from the first tuple and add `1`

PR #613, opened 2017-12-07
for add http proxy support

xiaomingplus:
give elasticsearch js http proxy support.

you can call like:

```javascript
const client = new elasticsearch.Client({
    hosts: [{
        host:'http://xxx.elk.com',
        proxy:"http://127.0.0.1:8080",
        port:80
    }]
});
```

### DIFF
--- a/src/lib/connectors/http.js
+++ b/src/lib/connectors/http.js
@@ -106,13 +106,17 @@ HttpConnector.prototype.makeAgentConfig = function (config) {
 HttpConnector.prototype.makeReqParams = function (params) {
   params = params || {};
   var host = this.host;
-
+  var proxy = host.proxy;
+  var proxyUrlObj = null;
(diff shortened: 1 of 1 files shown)


xiaomingplus:
signed.should I re pull ?

karmi:
Hi @xiaomingplus, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-js/pull/613.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

xiaomingplus:
isn't xiaomingplus@qq.com  ? I think i leave a same address.

xiaomingplus:
I'm sorry. I have add the email to my github profile.thanks.

PR #643, opened 2018-03-13
Make unit tests pass in node 8

dominykas:
(added a comment on each change)

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -71,7 +71,10 @@ _.each([
 
   _['isArrayOf' + type + 's'] = function (arr) {
     // quick shallow check of arrays
-    return _.isArray(arr) && _.every(arr.slice(0, 10), check);
+    return _.isArray(arr) && _.every(arr.slice(0, 10), function (item) {
+
+      return check(item);
+    });
   };
 });
 
--- a/test/unit/specs/http_connector.js
+++ b/test/unit/specs/http_connector.js
@@ -434,7 +434,7 @@ describe('Http Connector', function () {
       con.request({
         body: body
       }, function (err, resp, status) {
-        expect(http.ClientRequest.prototype.setHeader.lastCall.args).to.eql(['Content-Length', 14]);
+        expect(http.ClientRequest.prototype.setHeader.args.find((arg) => arg[0] === 'Content-Length')).to.eql(['Content-Length', 14]);
         server.done();
         done();
       });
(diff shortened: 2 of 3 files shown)


test/unit/specs/http_connector.js
```diff
@@ -444,7 +444,10 @@ describe('Http Connector', function () {
       var con = new HttpConnection(new Host());
       var respBody = 'i should not be encoded';
       var server = nock('http://localhost:9200')
-      .matchHeader('accept-encoding', undefined)
+      .matchHeader('accept-encoding', function (v) {
```
dominykas: `nock@9` doesn't seem to do a match on `undefined` - changing to a matcher function.

test/unit/specs/http_connector.js
```diff
@@ -434,7 +434,7 @@ describe('Http Connector', function () {
       con.request({
         body: body
       }, function (err, resp, status) {
-        expect(http.ClientRequest.prototype.setHeader.lastCall.args).to.eql(['Content-Length', 14]);
+        expect(http.ClientRequest.prototype.setHeader.args.find((arg) => arg[0] === 'Content-Length')).to.eql(['Content-Length', 14]);
```
dominykas: `nock@9` seems to add the `host` header, so `content-length` is no longer the `lastCall`.

src/lib/utils.js
```diff
@@ -71,7 +71,10 @@ _.each([
 
   _['isArrayOf' + type + 's'] = function (arr) {
     // quick shallow check of arrays
-    return _.isArray(arr) && _.every(arr.slice(0, 10), check);
+    return _.isArray(arr) && _.every(arr.slice(0, 10), function (item) {
+
+      return check(item);
```
dominykas: `_.every()` calls the function with `(item, index)` - if we directly pass that through - node 8 crashes, because `isRegexp` check expects a single argument; the fix is to make sure we call `check()` with a single argument.

I do wonder, though, if all of the `isArrayOf...` stuff should be removed altogether, as the only time I see this used is a single `isArrayOfStrings()` call.

Also curious about the `slice()` - is that for performance reasons that this only checks the first 10 members of an array?

spalger: yeah, I was probably being overly ambitious when I first wrote these utils and thought I would use them for something... Feel free to delete anything that's not used and inline little helpers like `isArrayofString()` rather than keeping it in `lib/utils` if it's only used once.

Issue #630, opened 2018-02-15
Upgrade lodash to v4.17.5+


dominykas:
This is using `lodash@2.x`, which has a vulnerability: https://snyk.io/vuln/npm:lodash:20180130

I might revisit and PR this at some point later - dealing with other modules having the same problem in the meantime...

dominykas:
OK, having had a quick look now... I can probably devote some time to it, but I'd appreciate some guidance.

I'm unsure about how to get the whole test suite to run; it errors out with a weird error in node 8 (I did get the unit tests running on node 6). Does it need a specific environment? Is there a CI server anywhere?

As the version is pinned, and there's some overrides in `utils.js`, there's clearly some interesting history to be aware of.

Please let me know if there's any place/time we can talk!

dominykas:
Before I can proceed discussing this at work, I do have some questions on the process here...

1. I see some mentions of travis stuff, but I don't see it enabled anymore - is there any kind of CI behind this somewhere out of the public eye?
2. Which versions of node/browser are supposed to be supported? I see most of the code is ES5?
3. I've done a couple of lodash 2->3->4 migrations in codebases I'm familiar with... but what would _your_ estimate of how long it would take for this lib? blind guess is fine... just calibrating my bearings.
4. I know there's issues getting integration tests to run... but is there a summary of what it takes to do it? Even if they're flaky - they might still be useful...
5. As I can't get `grunt coverage` to run... what was the last known coverage? How much can I trust unit tests?
6. What's `master` branch for (`14.x` is the default...)?
7. What's your take on "unbundling" utils back into lodash and fs? (and whatever remains as utils)

Sorry if this is a lot :D like I said - just trying to scope out and understand how big of a task this is.

I don't mind updating any docs with anything I learn in getting things to run... just need some pointers before I start.

spalger:
> is there any kind of CI behind this somewhere out of the public eye?

Not at this time

> Which versions of node/browser are supposed to be supported? I see most of the code is ES5?

As of version 14.0 esjs only supports node 4+. Also as of version 14.0 browser builds are not currently being transpiled, which means that they probably don't have great browser support but that should be fixed by #645

> what would *your* estimate of how long it would take for this lib?

Things I'm considering when estimating:
 - I was pretty new to lodash when I first wrote this lib so I don't expect much esoteric feature usage
 - you are already familiar with the breaking changes that will need to be accounted for
[...]

PR #645, opened 2018-03-15
[webpack] use babel to ensure legacy browser compatibility

spalger:
Inspired by @dominykas question in #630 I decided to look into browser version support and realized that there wasn't any transpiling happening to make sure that browser client is compatible with legacy browsers like IE 11. This fixes that by transpiling the bundles with the `@babel/preset-env` targeting `last 2 versions, > 5%, Safari 7, // for PhantomJS support`, which currently translates to:

http://browserl.ist/?q=last+2+versions%2C+%3E+5%25%2C+Safari+7

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,10 +45,13 @@
     ]
   },
   "devDependencies": {
+    "@babel/core": "^7.0.0-beta.41",
+    "@babel/preset-env": "^7.0.0-beta.41",
     "@spalger/eslint-config-personal": "^0.4.0",
     "aliasify": "^2.1.0",
     "async": "~0.8.0",
-    "babel-eslint": "^6.0.4",
+    "babel-eslint": "^8.2.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -107,6 +110,9 @@
     "lodash.isempty": "^4.4.0",
     "lodash.trimend": "^4.5.1"
   },
+  "resolutions": {
+    "grunt-webpack/deep-for-each": "https://github.com/spalger/js-deep-for-each/releases/download/v2.0.2-fix-missing-lodash-dep/deep-for-each-2.0.2.tgz"
+  },
```
archanid: What's our story on this dependency, is this something we can contribute a fix for?

spalger: Aye, https://github.com/moxystudio/js-deep-for-each/pull/23

Once it is released, and assuming that `grunt-webpack` picks up the new version, we can remove this.

PR #660, opened 2018-05-08
Upgrade to lodash v4

### LINKED ISSUE #630: Upgrade lodash to v4.17.5+

dominykas:
Closes #630 

How I approached this:

1. go through each [breaking change of lodash v2 to v3](https://github.com/lodash/lodash/wiki/Changelog#v300) - check every usage of a function that changed (i.e. search by function name)
2. same for [lodash v3 to v4](https://github.com/lodash/lodash/wiki/Changelog#v400)
3. go through every `require` of `src/lib/utils.js` and split up the `_` vs `utils` usage
4. go through every usage of `_.` and `_(` to verify that `thisArg` is no longer used (breaking change in v4)


package.json
```diff
@@ -106,8 +106,8 @@
     "agentkeepalive": "^3.4.1",
     "chalk": "^1.0.0",
     "lodash": "2.4.2",
+    "lodash-2": "4.17.4",
```
cstlaurent: Sorry for intruding, but what is this lodash-2 package? With 1 version deployed a year ago and having two weekly downloads on npm, it looks quite fishy to me. I also cannot find any reference to lodash-2 in the official GitHub org. I wouldn't want that appearing in my code base...

cstlaurent: Oh, I understand, I thought that `lodash-2` was to pin a lodash 2.x version and the main `lodash` package would be upgraded to latest. I guess the the names and versions makes it confusing 😃. Anyways, great work!

dominykas:
OK, this is now done, `lodash-2` is cleaned out.

Unit tests are passing.

I'll test this with our internal system, but otherwise I'm not sure how to be certain that I did a good job...

src/lib/client.js
```diff
@@ -92,8 +93,7 @@ function Client(config) {
         Log: require('./log'),
         Logger: require('./logger'),
         NodesToHost: require('./nodes_to_host'),
-        Transport: require('./transport'),
-        utils: require('./utils')
```
dominykas: Technically, removing this is a breaking change? This doesn't seem to be used internally - but it might be used by plugins?

Even not removing it is probably a breaking change, as `utils` no longer includes `lodash` methods and even if it were to include them - lodash v4 methods are breaking, so I'm not sure on the best course of action here?

dominykas:
When I run `grunt run:generate`, I get a `Error: Multiple URLS with the same signature detected for indices.putMapping` - but I get the same on `14.x` branch, so assuming it's not something I broke?

dominykas:
Unit tests in our internal system passed (against ES 6.1.3) - I'll check E2E tomorrow or Monday. Of course we don't use _that_ much of the functionality, but still...

spalger:
🎉 🎉  Awesome!

I'll take a deeper look Monday and do what I can do double check your efforts. Thank you so much @dominykas!!

dominykas:
Can confirm no issues E2E using this version on our internal system.

synhershko:
:+1:  thanks guys!

dominykas:
Thanks! Now 🤞I didn't break anything :)

PR #677, opened 2018-06-30
Fix: Serialization does not handle non-enumerable properties in Node <5.x

### LINKED ISSUE #504: Serialization does not handle non-enumerable properties in Node versions < 5.x
I'm observing this behavior in Node v4.5.0, but it is presumably affecting all versions of Node < v5.x because the underlying bug is in the V8 JSON.stringify implementation.

The problem arises when indexing a document that is an object with non-enumerable properties in some edge cases of prototypal inheritance conflicts.
Because the serialize function [here](https://github.com/elastic/elasticsearch-js/blob/master/src/lib/serializers/json.js#L23) utilizes `replacer` and `spaces` even if they are `undefined`. Even if the values are not `undefined`, this behavior is still exhibited. But my proposed solution mitigates to risk to only those cases relying upon `replacer` and `spaces`

```js
// Node v6.9.0
var Thing = function() {
    this.name = 'pretend that I am serialized';

    Object.defineProperty(this, 'schema', {
        value: {
            schema: 'this should not be serialized'
        }
    });
};
Thing.prototype.schema = 'Overwriting the non-enum to trigger JSON bug';

var thing = new Thing();

JSON.stringify(thing, undefined, undefined);
// '{"name":"pretend that I am serialized"}'
```

```js
// Node v4.5.0
var Thing = function() {
    this.name = 'pretend that I am serialized';

    Object.defineProperty(this, 'schema', {
        value: {
            schema: 'this should not be serialized'
        }
    });
};
Thing.prototype.schema = 'Overwriting the non-enum to trigger JSON bug';

var thing = new Thing();

JSON.stringify(thing, undefined, undefined);
// '{"name":"pretend that I am serialized","schema":{"schema":"this should not be serialized"}}'
```

Proposed solution. Update [this](https://github.com/elastic/elasticsearch-js/blob/master/src/lib/serializers/json.js#L23) to this:
```js
if (replacer || spaces) {
    return JSON.stringify(val, replacer, spaces);
}
else {
    return JSON.stringify(val);
}
```

As a bonus, this also comes with slight performance increases. Please see discussions in the Hapi.js project on this topic [here](https://github.com/hapijs/hapi/pull/3014) and [here](https://github.com/hapijs/hapi/blob/10bcdaaa9c24c9d2520455055e465d6ef67b206b/lib/response.js#L573-L580)

mirkojotic:
I just implemented suggestion by @WesTyler and added some unit tests to ensure it is working as it should. All other tests are passing. Closes #504

### DIFF
--- a/src/lib/serializers/json.js
+++ b/src/lib/serializers/json.js
@@ -20,7 +20,11 @@ Json.prototype.serialize = function (val, replacer, spaces) {
       return val;
     case 'object':
       if (val) {
-        return JSON.stringify(val, replacer, spaces);
+        if (replacer || spaces) {
+          return JSON.stringify(val, replacer, spaces);
+        } else {
+          return JSON.stringify(val);
+        }
       }
     /* falls through */
     default:
--- a/test/unit/specs/json_serializer.js
+++ b/test/unit/specs/json_serializer.js
@@ -39,6 +39,24 @@ describe('JSON serializer', function () {
         ser.serialize(thing);
       }).to.throwError();
     });
+
+    it('utilizes replacer or spaces if passed', function () {
+      sinon.spy(JSON, 'stringify');
+      var ser = makeSerializer();
+      var thing = { name: 'thing' };
+      ser.serialize(thing, null, 2);
+      expect(JSON.stringify.withArgs(thing, null, 2).calledOnce).to.be(true);
+      JSON.stringify.restore();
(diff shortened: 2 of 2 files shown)


spalger:
LGTM! Thanks @mirkojotic

spalger:
Just released this in v15.1.1

Issue #547, opened 2017-05-19
Ping spec requestTimeout overrides client requestTimeout


jbudz:
It looks like the requestTimeout at https://github.com/elastic/elasticsearch-js/blob/master/src/lib/apis/master.js#L5557 overrides the requestTimeout when initializing a client.

Not sure if it's intentional, if it is we should to override this in the kibana health check.

tylersmalley:
Test:

```
const elasticsearch = require('elasticsearch');

const client = new elasticsearch.Client({
  host: 'localhost:9200',
  log: 'trace',
  requestTimeout: 30000,
  pingTimeout: 30000
});

client.ping();
```

Log:
```
Elasticsearch INFO: 2017-05-19T16:50:41Z
  Adding connection to http://localhost:9200/

Elasticsearch DEBUG: 2017-05-19T16:50:42Z
  starting request {
    "method": "HEAD",
    "requestTimeout": 3000,
    "castExists": true,
    "path": "/",
    "query": {}
  }


Elasticsearch TRACE: 2017-05-19T16:50:42Z
  -> HEAD http://localhost:9200/

  <- 200


Elasticsearch DEBUG: 2017-05-19T16:50:42Z
  Request complete
```

bohyun-e:
@jbudz @tylersmalley Any updates on this issue?

amiram:
This bug seems simple. There is a hard-coded value for timeout in ping function in each api. For example here: https://github.com/elastic/elasticsearch-js/blob/master/src/lib/apis/6_0.js#L5515
Is it there for a reason?

PR #679, opened 2018-07-02
[ping()] determine ping timeout at runtime

jbudz:
Currently ping does not respect defaults set on client instantiation.   The only way to override is by setting request parameters. 

This sets the ordering from:
1) requestTimeout param
2) 3000

to:
1) pingTimeout param
1) requestTimeout param
1) instantiated pingTimeout
1) default pingTimeout

Closes https://github.com/elastic/elasticsearch-js/issues/547

### DIFF
--- a/src/lib/transport.js
+++ b/src/lib/transport.js
@@ -206,9 +206,8 @@ Transport.prototype.request = function (params, cb) {
 
   const pingRequest = params.path === '/' && params.method === 'HEAD';
   if (pingRequest) {
-    const pingParam = params.hasOwnProperty('pingTimeout') && params.pingTimeout;
     const requestParam = params.hasOwnProperty('requestTimeout') && params.requestTimeout;
-    requestTimeout = pingParam || requestParam || this.pingTimeout || requestTimeout;
+    requestTimeout = requestParam || this.pingTimeout;
   }
 
   params.req = {
--- a/src/lib/transport.js
+++ b/src/lib/transport.js
@@ -37,6 +37,7 @@ function Transport(config) {
 
   // setup requestTimeout default
   self.requestTimeout = config.hasOwnProperty('requestTimeout') ? config.requestTimeout : 30000;
+  self.pingTimeout = config.hasOwnProperty('pingTimeout') ? config.pingTimeout : 3000;
 
   if (config.hasOwnProperty('defer')) {
     self.defer = config.defer;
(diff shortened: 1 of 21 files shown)


jbudz:
I need to update tests still but want to run the approach by someone first.  I'm not a fan of this approach but it's the best spot I could find.  It's not totally clear to me the difference between client.abstract.ping and the spec ping, and the isPingRequest would preferably go by spec name.

tylersmalley:
Looks like it's still being set to 3000

jbudz:
oh on it - we have to re-generate the spec files, I left it out cuz they're generated but just realized it probably doesn't make testing easy.

src/lib/transport.js
```diff
@@ -203,6 +204,15 @@ Transport.prototype.request = function (params, cb) {
     requestTimeout = params.requestTimeout;
   }
 
+  const pingRequest = params.path === '/' && params.method === 'HEAD';
+  if (pingRequest) {
+    const pingParam =
```
tylersmalley: nit: I don't think it's necessary to split these lines up and IMO makes it easier to read

```
const pingParam = params.hasOwnProperty('pingTimeout') && params.pingTimeout;
const requestParam = params.hasOwnProperty('requestTimeout') && params.requestTimeout;

requestTimeout = pingParam || requestParam || this.pingTimeout || requestTimeout;
```

src/lib/transport.js
```diff
@@ -203,6 +204,15 @@ Transport.prototype.request = function (params, cb) {
...
+      params.hasOwnProperty('pingTimeout') && params.pingTimeout;
+    const requestParam =
+      params.hasOwnProperty('requestTimeout') && params.requestTimeout;
+    requestTimeout = pingParam || requestParam || this.pingTimeout || requestTimeout;
```
tylersmalley: I believe the last `requestTimeout` should be `this.requestTimeout` which is set on line 39. Otherwise, the `requestParam` and the `requestTimeout` are the same value, additionally, we will never get to the `requestTimeout` since we default the `pingTimeout` to 3000.

What about ditching the requestTimeout from this altogether. Use whats passed in the params or default to the config?

```
const isPingRequest = params.path === '/' && params.method === 'HEAD'

if (isPingRequest) {
  requestTimeout = params.hasOwnProperty('pingTimeout') ? params.requestTimeout : this.pingTimeout;
}
```

Actually, how do I even pass in the pingTimeout in the params?

src/lib/transport.js
jbudz: 👍 that was an autofix done by prettier, but it's not enabled here so i'll revert

src/lib/transport.js
jbudz: yeah this gets a little confusing :).  requestTimeout is set  on line 152 to this.requestTimeout.  it's potentially overridden on line 203 by the requestTimeout param.  we don't want to use this.requestTimeout because it falls back to the global default of 30000ms or instantiated timeout, when we want to fallback to the global pingTimeout

technically i can remove it entirely,, because pingTimeout has a default now.  would that work?

pingParam is function param OR a default included in the generated spec
```
client.ping({ pingTimeout: 3000 })
```

this.pingTimeout is the global default of 3000ms, defined by the client  OR overridden on client instantiation.

so we get an ordering of
1) pingTimeout param on execute
1) requestTimeout param on execute
1) pingTimeout param on client instantiation
1) global default of 3000ms

jbudz:
@spalger if you get a moment can you take a look?  No pressure on it

tylersmalley:
Maybe @delvedor can review?

PR #708, opened 2018-10-15
Fix memory leaks when making requests and logging

### LINKED ISSUE #707: Many API requests causes memory leak
I have discovered a memory leak using the latest version of `elasticsearch-js@v15.1.1` and `v14.2.2`. The leak occurs when making repeated requests against elasticsearch, in my case, I am creating a bunch of records in elasticsearch.

I have created a [repo](https://github.com/peterdemartini/elasticsearch-js-leak) with a test and a script to show the effects of the leak.

peterdemartini:
I created a test [repo](https://github.com/peterdemartini/elasticsearch-js-leak) that shows there are memory leaks in `elasticsearch-js`. After taking a few heap snapshots and finding the leaks, I fixed a few of the problems. Most of the leaks were caused by the use of `_.bind`, holding on references longer than needed, and creating sparse arrays by continually mutating them.

I used a couple of es6 features in my changes, like the arrow functions, but this doesn't seem to be used elsewhere in within project, let me know if I should change this.

After applying my changes to the test project, I saw **10x** decrease in memory gain, see the [results](https://github.com/peterdemartini/elasticsearch-js-leak/blob/master/RESULTS.md).

Resolves #707.

### DIFF
--- a/src/lib/client.js
+++ b/src/lib/client.js
@@ -57,11 +57,11 @@ function Client(config) {
 
     this.transport = new Transport(config);
 
-    _.each(EsApiClient.prototype, _.bind(function (Fn, prop) {
+    _.each(EsApiClient.prototype, (Fn, prop) => {
       if (Fn.prototype instanceof clientAction.ApiNamespace) {
         this[prop] = new Fn(this.transport, this);
       }
-    }, this));
+    });
 
     delete this._namespaces;
   }
--- a/src/lib/client_action.js
+++ b/src/lib/client_action.js
@@ -62,7 +62,7 @@ function makeFactoryWithModifier(modifier) {
         return exec(this.transport, spec, _.clone(params), cb);
       } catch (e) {
         if (typeof cb === 'function') {
-          utils.nextTick(cb, e);
+          process.nextTick(cb, e);
         } else {
           var def = this.transport.defer();
           def.reject(e);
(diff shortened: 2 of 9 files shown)


delvedor:
Hi! Thank you for the great work!
We are currently in the process of redesigning the client from scratch (the work is still private), soon we will publish a pr with the new code.

alanhoff:
@delvedor is there an ETA for the rewrite to be public?

delvedor:
@alanhoff not yet, but we will release the first alpha this week :)

Remexllee:
facing memory leak also, would like to know when will release this fix?

tylersmalley:
jenkins, test it

peterdemartini:
What is the status of this PR @davalapar?

tusbar:
Hey @spalger – can something be done about this? I’m seeing crazy memory leaks.

laukaichung:
Has `@elastic/elasticsearch` fixed this memory issue? I don't see any difference. The memory usage increases 100mb+ when making a search request. It never gets released.

jaydp17:
Try out
```js
const client = new elasticsearch.Client({
  ...
  keepAlive: false,
});
```
Even after this, I'm still seeing some memory leaking after every search query, but the growth rate is small enough that for now, it'll work.

Also, I'm not sure if this helps to debug the issue, but I'm using webpack to bundle the code, maybe that's somehow contributing to the issue.

_P.S. I'm on `v15.4`_

laukaichung:
Would you like to try `@elastic/elasticsearch`? I'd like to see if the memory issues really get solved. I'm using it but I don't see any difference. The memory still spikes up when making requests to the elasticsearch server.

delvedor:
Hello!
If you are having a problem with `@elastic/elasticsearch` can you open a new issue with an example and instructions to reproduce it?
Thanks!

PR #723, opened 2018-11-02
WIP: clients-ci

delvedor:
Initial work for the `clients-ci`.

Notes:
- the integration testing suite is using `http://localhost:9200` as default host, should I use `http://elasticsearch:9200`
- same as above for `wait-cluster.sh` script

Feedbacks are welcome :)


spalger:
> the integration testing suite is using http://localhost:9200 as default host, should I use http://elasticsearch:9200

I think localhost is a perfectly reasonable default that the docker-ized version of the tests can override, so I would not default to elasticsearch:9200

.ci/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:${NODE_JS_VERSION}
```
fxdgear: missing the `ARG` section at the top. 

```
ARG NODE_JS_VERSION=8
FROM node:${NODE_JS_VERSION}
```

.ci/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+# Create app directory
+WORKDIR /usr/src/app
+
+# Install app dependencies
+COPY package*.json ./
```
fxdgear: So there's an issue here with this line, and there's a couple ways we can solve it. 

1. move the `docker-compose.yml` up one directory to the root of the repo. 
2. Adjust the docker-compose build context to reference the correct location. 

the problem right now is that `Dockerfile` is trying to copy a file that doesn't exist. (it exists one directory up).

package.json
```diff
@@ -19,7 +19,8 @@
     "lint": "standard",
     "lint:fix": "standard --fix",
     "generate": "node scripts/run.js",
-    "elasticsearch": "docker run --rm -e \"node.attr.testattr=test\" -e \"path.repo=/tmp\" -e \"repositories.url.allowed_urls=http://snapshot.*\" -p 9200:9200 docker.elastic.co/elasticsearch/elasticsearch:6.4.0"
+    "elasticsearch": "docker run --rm -e \"node.attr.testattr=test\" -e \"path.repo=/tmp\" -e \"repositories.url.allowed_urls=http://snapshot.*\" -p 9200:9200 docker.elastic.co/elasticsearch/elasticsearch:6.4.0",
+    "ci": "./scripts/wait-cluster.sh && npm test && npm run test:integration"
```
fxdgear: So trying to run this locally failed cause the `wait-cluster.sh` script is refrencing `localhost:9200` vs `http://elasticsearch:9200`. 

here's what I did and things work... not sure how you use this script otherwise... so maybe we need to check if `TEST_ES_SERVER` env var exists first and maybe default it? 

```bash
#!/bin/bash

attempt_counter=0
max_attempts=5
url="${TEST_ES_SERVER}/_cluster/health?wait_for_status=green&timeout=50s"

echo "Waiting for Elasticsearch..."
echo $url
while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' --max-time 55 "$url")" != "200" ]]; do
  if [ ${attempt_counter} -eq ${max_attempts} ];then
    echo "\nCouldn't connect to Elasticsearch"
    exit 1
  fi

  printf '.'
  attempt_counter=$(($attempt_counter+1))
  sleep 5
done

echo "\nReady"
```

fxdgear:
> I think localhost is a perfectly reasonable default that the docker-ized version of the tests can override, so I would not default to elasticsearch:9200

I think it's ok to default to localhost:9200 as well, so lets adjust the script to pull in the ENV variable `TEST_ES_SERVER`. 

We can default `TEST_ES_SERVER` to `localhost:9200` if the ENV doesn't exist, but the script should be using `TEST_ES_SERVER` as the root of the url to curl against.

.ci/Dockerfile
delvedor: I went for option 2 :)

fxdgear:
:boom:

delvedor:
Jenkins retest this please

fxdgear:
the broken ci test are fixed now! :)

PR #727, opened 2018-11-12
Document serializer configuration option

### LINKED ISSUE #695: Stable JSON stringify to avoid cache misses
It looks like this lib [is using JSON.stringify](https://github.com/elastic/elasticsearch-js/blob/master/src/lib/serializers/json.js#L24-L26) on requests. Per [the elasticsearch query cache docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/shard-request-cache.html#_cache_key):
> if the JSON changes — for instance if keys are output in a different order — then the cache key will not be recognised.

This means that if we make requests to the elasticsearch client with objects where the keys are not created in the exact same order, we can't take advantage of the request cache. My unstanding is that these two would not be the same from the cache's perspective:

```js
await client.search({
  index: 'twitter',
  type: 'tweets',
  body: {
    query: {
      match: {
        body: 'elasticsearch'
      }
    }
  }
})
```

vs

```js
await client.search({
  type: 'tweets',
  index: 'twitter',
  body: {
    query: {
      match: {
        body: 'elasticsearch'
      }
    }
  }
})
```

If we switch to any of [these libs](https://www.npmjs.com/search?q=stable%20stringify), we could make sure the request cache is hit more often. If it's much slower, maybe we could make it a config option for the client (e.g. `stableStringify: true`)?

daedalus28:
Closes #695 

This PR adds documentation for the `serializer` configuration option. From the commit history, it looks like this file _might_ be generated, so please let me know if there's anywhere else I need to update. I'm not familiar with the asciidoc format, but I tried to follow the doc patterns I saw in the rest of the file so I hope it's correct.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -276,6 +276,31 @@ var client = new elasticsearch.Client({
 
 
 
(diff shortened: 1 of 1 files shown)


daedalus28:
Just signed the CLA

spalger:
👍 thanks!

PR #740, opened 2018-12-06
CI: test both oss and platinum APIs

delvedor:
Initial work to test both oss and platinum APIs.


.ci/docker-compose.yml
```diff
@@ -18,20 +18,67 @@ services:
...
+          NODE_JS_VERSION: ${NODE_JS_VERSION:-8}
+    environment:
+      - "TEST_ES_SERVER=http://elastic:secret@elasticsearch-commercial:9201"
+      - "ELASTIC_PASSWORD=secret"
+    volumes:
+      - ..:/usr/src/app
+      # This will mount the node_modules directory
+      # to the host machine using the buildtime directory.
+      - /usr/src/app/node_modules
+      - esvol:/tmp
```
fxdgear: don't you need these two lines in the `client-oss`?

.ci/run-tests
```diff
@@ -10,4 +10,5 @@
 #   - $NODE_JS_VERSION
 #
 
-ELASTICSEARCH_VERSION=${ELASTICSEARCH_VERSION} NODE_JS_VERSION=${NODE_JS_VERSION} docker-compose -f .ci/docker-compose.yml run client
+ELASTICSEARCH_VERSION=${ELASTICSEARCH_VERSION} NODE_JS_VERSION=${NODE_JS_VERSION} docker-compose -f .ci/docker-compose.yml run client-oss
+ELASTICSEARCH_VERSION=${ELASTICSEARCH_VERSION} NODE_JS_VERSION=${NODE_JS_VERSION} docker-compose -f .ci/docker-compose.yml run client-commercial
```
fxdgear: I'm not sure how I feel about this. 
one one hand it works... 
one another the jenkins worker isn't really isolating the test. 

Maybe we can do some work with the test matrix? To add a another "element" to the matrix?

My point being if a test fails in the commertial api but dos not fail in the oss lib then we'll get a failure in jenkins and it won't be obviously clear which one failed...

scripts/es-commercial-docker.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
fxdgear: do we need this?

can we not just alter the other one to accept CLI paremeters, and if the `elastic_password` is set then we append all the xpack stuff otherwise we execute as normal?

delvedor: I'm using it in my machine to run quick test :)

.ci/run-tests
delvedor: This morning @karmi made the same point and I agree completely.
How can we achieve that?

.ci/docker-compose.yml
delvedor: Sorry, I fear I'm missing your point. Can you rephrase?

PR #746, opened 2018-12-20
Add JJBB jobs to .ci

jonahbull:
This adds the Jenkins jobs definitions to the repo so they can be
managed and updated here.


delvedor:
No worries!

> SI can just open another PR that's based on next if that'd be easier?

Probably yes :)

jonahbull:
The action moves to #749

PR #748, opened 2018-12-21
Request parameters typings

delvedor:
This pr adds the automated generation of the request parameters typings 🎉 
Feedbacks are welcome!

cc @Mpdreamz @timroes @sqren


api/requestParams.d.ts
```diff
@@ -0,0 +1,1895 @@
...
+  h?: string | string[];
+  help?: boolean;
+  s?: string | string[];
+  v?: boolean;
+}
+
+export interface CatAllocation extends Generic {
+  node_id?: string | string[];
+  format?: string;
+  bytes?: 'b' | 'k' | 'kb' | 'm' | 'mb' | 'g' | 'gb' | 't' | 'tb' | 'p' | 'pb';
```
Mpdreamz: Create type `Bytes` enum and reference that? as well as for the other inline unions.

delvedor: I had a chat with @timroes, and we agreed that since there is not a real benefit on generalizing this *(the autocompletion would still work)* and the code generation complexity will increase a lot, for now, I'll leave this as it is :)

timroes:
Haven't done an in depth analysis of the generation code, but talked about the actual types and exporting, and that looks good to me.

PR #749, opened 2018-12-21
Add JJBB jobs to .ci

jonahbull:
This adds the Jenkins jobs definitions to the repo so they can be
managed and updated here.

### DIFF
--- a/.ci/jobs/elastic+elasticsearch-js+next.yml
+++ b/.ci/jobs/elastic+elasticsearch-js+next.yml
@@ -1,0 +1,11 @@
+---
+- job:
+    name: elastic+elasticsearch-js+next
+    display-name: 'elastic / elasticsearch-js # next'
+    description: Testing the elasticsearch-js next branch.
+    parameters:
+    - string:
+        name: branch_specifier
+        default: refs/heads/next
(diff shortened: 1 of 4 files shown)


jonahbull:
Here you go, @delvedor, hopefully this works out better than my last attempt! 😅

.ci/jobs/elastic+elasticsearch-js+next.yml
```diff
@@ -0,0 +1,11 @@
+---
+- job:
```
delvedor: What is the trigger for this job?

jonahbull: It'll inherit the default `github` trigger from `defaults.yml` when JJBB merges the jobs and defaults.

delvedor: Great!
And the `github` trigger means that it will run a build for every commit in the branch?

jonahbull: Yep, that's right.

These files will produce the same behavior as the jobs that currently exist on `clients-ci`.

PR #756, opened 2019-01-23
[connectors/http] prevent errors caused by multi-byte characters in paths

spalger:
Since https://github.com/nodejs/node/commit/b961d9fd83, `http.request()` will throw an error if multi-byte characters are used in the URL path, which is not currently being caught and passed back to the callback.

This may be true in a lot of places, but such is the nature of callbacks.

### DIFF
--- a/test/unit/specs/http_connector.js
+++ b/test/unit/specs/http_connector.js
@@ -232,14 +232,12 @@ describe('Http Connector', function () {
     });
 
     it('calls back with error when http.request throws an error', function (done) {
-      var con = new HttpConnection(new Host('http://google.com'));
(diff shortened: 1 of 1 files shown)


LeeDr:
@spalger does this PR fix it just so Kibana doesn't crash on multi-byte chars but it still fails?  
Or does this fix it so we can successfully use multi-byte chars in index pattern names?

spalger:
@LeeDr this does not allow us to use multi-byte chars in Kibana, but it will properly forward the errors to the callback rather than crashing the process. See https://github.com/elastic/kibana/pull/29210 for the other part of the solution.

PR #761, opened 2019-01-29
Unknown parameters handling

delvedor:
Send unknown parameters as query parameters and generate a warning.
The generated warning will be `Client - Unknown parameter: "unknown_key", sending it as query parameter`.


delvedor:
I've also refactored the code generation, now the final code is more readable and with the same performances!

PR #768, opened 2019-02-20
Child client support

delvedor:
With this pr we introduce the `client.child` API, which returns a new client instance that shares the connection pool with the parent client.
This feature can be handy if you need to have multiple client instances with different configurations, but with a shared connection pool.

Example:
```js
const child = client.child({
  headers: { 'x-foo': 'bar' },
  requestTimeout: 1000
})

client.info(console.log)
child.info(console.log)
```

**Open questions:**
- Currently, the event emitter is shared between the parent and the child(ren), is it ok?
- Currently, if you extend the parent client, the child client will have the same extensions, while if the child client adds an extension, the parent client will not be extended. Is it ok?

**Caveats:**
- You can override *any* option except for the connection pool specific options (`ssl`, `agent`, `pingTimeout`, `Connection`, and `resurrectStrategy`).
- You can't specify a new `Connection` class.
- If you call `close` in any of the parent/child clients, every client will be closed.

*Note: the `nodeFilter` and `nodeSelector` options are now `Transport` options and no longer `ConnectionPool` options.*

### DIFF
--- a/docs/child.asciidoc
+++ b/docs/child.asciidoc
@@ -1,11 +1,11 @@
 = Creating a child client
 
 There are some use cases where you may need multiple instances of the client. You can easily do that by calling `new Client()` as many times as you need, but you will lose all the benefits of using one single client, such as the long living connections and the connection pool handling. +
-To avoid this problem the client offers a `child` API, which returns a new client instance that shares the connection pool with the parent client. +
+To avoid this problem the client offers a `child` API, which returns a new client instance that shares the connection pool with the parent client. +
 
 NOTE: The event emitter is shared between the parent and the child(ren), and if you extend the parent client, the child client will have the same extensions, while if the child client adds an extension, the parent client will not be extended.
 
-You can pass to the `child` every client option you would pass to a normal client, but the connection pool specific options (`ssl`, `agent`, `pingTimeout`, `Connection`, and `resurrectStrategy`).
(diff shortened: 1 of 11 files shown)


epixa:
This seems to be exactly what I was hoping for. For what it's worth, I think both of your open questions are OK.

> If you call `close` in any of the parent/child clients, every client will be closed.

Given Kibana's use case for using this on every incoming request, what's the appropriate way for it to clean up after each child client at the end of the request?

delvedor:
> > If you call `close` in any of the parent/child clients, every client will be closed.

> Given Kibana's use case for using this on every incoming request, what's the appropriate way for it to clean up after each child client at the end of the request?

In theory, you don't need to. Every component of the Client except the `ConnectionPool` and its `Connection`(s) are stateless. This means that as soon as you stop using it, the garbage collector will clean it eventually, as it will be doing in any case for the route handlers.
The most important thing is not to create multiple closures that will increase the memory footprint.
The only element that will always be "allocated" is the initial client, since it holds the connection pool reference, while every other child client will use the parent pool.

Anyhow, I'm running some test to verify the memory impact for this solution.
The only problem so far is that when you create a child, you lose the benefits of the API methods lazy loading.

delvedor:
I'll merge this as is, the impact of the lazy loader is very low, and we can think about a different solution in the future if it becomes a problem.
Thank you for the review!

PR #778, opened 2019-03-15
Remove duplicate

cumom:
I don't think a duplicate parameter description is needed for @param type. Correct me if I am wrong.


delvedor:
Hello!
The bulk API allows you to send the `type` parameter in two different places, url and querystring. This is the reason why is duplicated in the documentation.
I've opened a pr that will help in this case.

Thank you for reporting!

PR #793, opened 2019-03-27
Make legacy docs asciidoctor compatible

nik9000:
This modifies the legacy client documentation so that it is is
compatible with Asciidoctor. With these changes asciidoctor renders the
same way as asciidoc, module spacing and a couple of auto-generated ids.
I'm willing to ignore the auto-generated anchors because these are
legacy docs.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -16,7 +16,7 @@ WARNING: Due to the complex nature of the configuration, the config object you p
 [[config-options]]
 === Config options
 [horizontal]
-`host or hosts`[[config-hosts]]::
+[[config-hosts]]`host or hosts`::
 `String, String[], Object[]` -- Specify the hosts that this client will connect to. If sniffing is enabled, or you call `client.sniff()`, this list will be used as seeds to discover the rest of your cluster.
 +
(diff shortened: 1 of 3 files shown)


lcawl:
I tried building with

> build_docs --asciidoctor --doc elasticsearch-js/docs/index.asciidoc --chunk 1
> build_docs --asciidoctor --doc elasticsearch-js/docs/index.asciidoc 
> docbldejs

... and they all seemed fine

docs/configuration.asciidoc
```diff
@@ -16,7 +16,7 @@ WARNING: Due to the complex nature of the configuration, the config object you p
 [[config-options]]
 === Config options
 [horizontal]
-`host or hosts`[[config-hosts]]::
+[[config-hosts]]`host or hosts`::
```
spalger: This file is generated from https://github.com/elastic/elasticsearch-js/blob/legacy-doc/scripts/generate/templates/configuration_docs.tmpl, so we'll need to make sure and update that template and run `node scripts/generate` to update this file.

delvedor: Thanks! I'll take care of this.

PR #800, opened 2019-03-29
Docs: Handle more funny cases from ES api spec

nik9000:
Handles a few "special" doc urls from ES's api spec files. These are
only a problem in 6.x but we'd like to keep the generateDocs script the
same across all branches for easier backporting.


nik9000:
Thanks for reviewing @delvedor! Sorry this is such a mess....

delvedor:
> Thanks for reviewing @delvedor! Sorry this is such a mess....

No worries, thank you for all the help!

Issue #809, opened 2019-04-09
Support for setting a custom http.agent (similar old createNodeAgent)


endymonium:
My use case is to add the missing proxy capabilities to the node.http/s library.

delvedor:
I've opened https://github.com/elastic/elasticsearch-js/pull/810, take a look! :)

delvedor:
Closed in #810.

Issue #808, opened 2019-04-09
Not able to add custom http.agent (createNodeAgent missing?)


endymonium:
Hello,

in the past there existed a createNodeAgent argument, which you could use to add a custom htt.agent. That was removed, isn't it? Are there any alternatives?

I need to add proxy support to the the node.http library ...

Thanks,
Jan

delvedor:
Hello!
Are you using the new client? (the one you install with `npm i @elastic/elasticsearch`)
If so, you can't pass a custom http agent at the moment, but it is a good feature request!

endymonium:
Yes, new client. I'll resubmit it as a fututre request then :)

PR #810, opened 2019-04-09
Custom http agent support

delvedor:
With this pr we add the support for a custom http(s) agent passed as a parameter instead of the http agent options.
In any case, you will use the same `agent` configuration option.

```js
const { Agent } = require('http')

const client = new Client({
  agent: new Agent({ ... })
})

client.info(console.log)
```

Related: https://github.com/elastic/elasticsearch-js/issues/808 https://github.com/elastic/elasticsearch-js/issues/809

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -90,8 +90,20 @@ _Default:_ `false`
 _Default:_ `null`
 
 |`agent`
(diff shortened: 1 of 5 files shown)


endymonium:
Tested it locally, looks good, thanks 👍 

```
var proxyingAgent = require('proxying-agent').create(
      'http://proxy:8080', 
      'https://<redacted>.eu-central-1.aws.cloud.es.io');
const client = new Client({ 
      node: 'https://<redacted>.aws.cloud.es.io',
      agent: proxyingAgent
})
```

PR #813, opened 2019-04-09
Improve typings

delvedor:
With this change, it's now easier to use TypeScript with the client.
The ApiResponse now accepts a [generics](https://www.typescriptlang.org/docs/handbook/generics.html) that defaults to `any`, same for every API method that might need a body.

The documentation has been updated accordingly.

I would like to thank @fox1t for the suggestions :)

### DIFF
--- a/docs/typescript.asciidoc
+++ b/docs/typescript.asciidoc
@@ -3,58 +3,80 @@
 
 The client offers a first-class support for TypeScript, since it ships the type definitions for every exposed API.
 
-While the client offers type definitions for Request parameters, Request bodies and responses are shipped with `any` because there is not an official spec that defines them, so we cannot make guarantees over them (but since they are shipped with `any`, you can easily override them with your own typing definitions).
+While the client offers type definitions for Request parameters, Request bodies and responses are shipped with https://www.typescriptlang.org/docs/handbook/generics.html[generics] (with a default type of `any`) because there is not an official spec that defines them, so we cannot make guarantees over them.
 
 NOTE: If you are using TypeScript you will be required to use _snake_case_ style to define the API parameters instead of _camelCase_. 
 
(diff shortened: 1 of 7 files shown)


fox1t:
LGTM

docs/typescript.asciidoc
```diff
@@ -3,58 +3,146 @@
...
+  skipped: number;
+}
+
+interface Explanation {
+  value: number;
+  description: string;
+  details: Explanation[];
+}
+
+interface SearchResponse<T> {
```
DASPRiD: I'm wondering, why can't this part be exported by this client library directly? This seems generic enough.

test/types/index.ts
```diff
@@ -94,6 +95,40 @@ client.index({
...
+// Dewfine the interface of the search response
+interface SearchResponse<T> {
+  hits: {
+    hits: Array<{
+      _source: T;
+    }>
+  }
+}
+
+// Define the intefrace of the source object
```
DASPRiD: "interface", not "intefrace" ;)

docs/typescript.asciidoc
delvedor: Because there is not an official spec from which I can generate the response types, and it will make no sense export the types just for the search response.

onehorsetown:
Shouldn't `ApiMethod` be typed like this, so if you are using promises, you don't have to cast anything:

```
declare type callbackFn<T> = (
  err: Error | null,
  result: ApiResponse<T>
) => void;

interface ApiMethod<T, R = any> {
  (params: T, options?: TransportRequestOptions): Promise<ApiResponse<R>>;
  (params: T, callback: callbackFn<R>): void;
  (params: T, options: TransportRequestOptions, callback: callbackFn<R>): void;
}
```

Issue #817, opened 2019-04-15
Export `ClientOptions`


onehorsetown:
##  Regression Report (kind of)

Previously (6.7) the Typescript typings from DefinitelyTyped exported the client`ConfigOptions` - which is seems to have been renamed to `ClientOptions` in 7.x:

See:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/elasticsearch/index.d.ts#L99

versus:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/elasticsearch/index.d.ts#L99

I realize Elastic probably did not publish these types, so this technically an enhancement request, as opposed to a regression, depending on how you look at it.

## Last working version

Worked up to version: 6.7

Stopped working in version: 7.0

## Expected behavior

Please add an `export` here:

https://github.com/elastic/elasticsearch-js/blob/7.x/index.d.ts#L67

as in:

```
export interface ClientOptions { 
   ...
}
```

or export `ClientOptions` where you export everything else on line 562.

## Your Environment

- *node version*: 11.12.0
- `@elastic/elasticsearch` *version*: >=7.0.0
- *os*: MacOS
- *any other relevant information*

delvedor:
Closed in #819.

PR #819, opened 2019-04-17
Updated typings

delvedor:
Follow-up on suggestions from https://github.com/elastic/elasticsearch-js/issues/813 and https://github.com/elastic/elasticsearch-js/pull/817.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -27,7 +27,8 @@ import Transport, {
   TransportRequestParams,
   TransportRequestOptions,
   nodeFilterFn,
-  nodeSelectorFn
+  nodeSelectorFn,
+  TransportRequestCallback
 } from './lib/Transport';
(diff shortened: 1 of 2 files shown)


onehorsetown:
@delvedor 

Thanks for this PR.  One last suggestion.  

```
export interface RequestEvent<T = any> {
    body: any;
    ...
```

should be:

```
export interface RequestEvent<T = any> {
    body: T;
    ...
```

As it is now, it `RequestEvent` is not using the parameterized type `T`

https://github.com/elastic/elasticsearch-js/blob/c6ce0628220b34bf2393ab4fa73cd849ac58f7c7/lib/Transport.d.ts#L52

delvedor:
@onehorsetown good catch!!
Do you want to open a pull request? :)

PR #824, opened 2019-04-21
Remove duplication from reference.asciidoc

sawell121:
Duplicated type entry, I left the entry at the top given that index + type are often mentioned in close proximity.


delvedor:

PR #834, opened 2019-04-29
Improve observability

delvedor:
With this pr, we add support for the request id. With it, it will be easier correlate clients events and in the future enhance the client observability.

It can be fully customized in the configuration options (as well as in the child clients) and can be overridden with the specific request options.

@watson @Qard: can this be useful for you? What can I do to improve the client observability for Elastic APM?

Example:
```js
const client = new Client({
  node: 'http://localhost:9200'
  // generateRequestId: function () { ... }
})

client.on('request', (err, { meta }) => {
  console.log(meta.request.id)
})

client.on('response', (err, { meta }) => {
  console.log(meta.request.id)
})

client.info(console.log)
client.info({}, { id: 'custom' }, console.log)
```

/cc @tylersmalley @epixa @spalger


Qard:
It seems to be in pretty good shape from APM perspective. With the `request` and `response` events we can capture the start and end of a request easily, and we have the metadata to capture the path, body, host, etc.

Thanks for the notice. :)

delvedor:
I'm thinking if it is worth to add a `context` object instead, which will contain the request id and any more data you might be interested in.

Eg:
```js
const client = new Client({
  node: 'http://localhost:9200'
})

client.on('request', (err, { meta }) => {
  console.log(meta.request.context) // { id: 0, winter: 'is coming' }
  meta.request.context.star = 'wars'
})

client.on('response', (err, { meta }) => {
  console.log(meta.request.context) // { id: 0, winter: 'is coming', star: 'wars' }
})

client.info({}, { context: { winter: 'is coming' } }, console.log)
```

Furthermore, I'm wondering if `generateRequestId` should also support async code. It should be quite easy to do.

tylersmalley:
I like the idea of being able to add any arbitrary data to the context. Would you still be able to have a `generateRequestId` on the client to automatically create an id?

delvedor:
Yes, if you don't do anything, `context` will be `{ id: 'some-id' }`, and the id will be generated by `generateRequestId` (unless you override it).

delvedor:
I've done some more changes which should improve the observability:
- event client instance now has a name (which can be configured), this is very useful when working with child clients.
- In every request, you can provide a custom `context` object that you can access later.
- Now you can access request id and client name also in the `resurrect` event.

In code:
```js
const client = new Client({
  name: 'cluster-abc'
})

client.on('request', (err, { meta }) => {
  const { name, context, request } = meta
  console.log(name, context, request.id)
})

client.info({}, { context: { winter: 'is coming' }, id: 'custom-id' }, console.log)
```

tylersmalley:
👍 this is great.

docs/observability.asciidoc
```diff
@@ -0,0 +1,251 @@
...
+  connection: Connection;
+  name: string;
+  request: {
+    id: any;
+  };
+}
+----
+
+=== Correlation id
+Correlate events can be quite hard, especially if there are happening many events at the same time, the client offers you an automatic (and configurable) system to help you handle this problem.
```
tylersmalley:
```suggestion
Correlating events can be quite hard, especially if there are many events at the same time. The client offers you an automatic (and configurable) system to help you handle this problem.
```

docs/observability.asciidoc
```diff
@@ -0,0 +1,251 @@
+[[observability]]
+== Observability
+
+The client does not provide a default logger, but instead it offers an event emitter interfaces to hook into internal events, such as `request` and `response`.
+
+Correlate those events can be quite hard, especially if your applications has a large codebase with many events happening at the same time.
```
tylersmalley:
```suggestion
Correlating those events can be quite hard, especially if your applications have a large codebase with many events happening at the same time.
```

docs/observability.asciidoc
```diff
@@ -0,0 +1,251 @@
...
+[source,js]
+----
+const { Client } = require('@elastic/elasticsearch')
+const client = new Client({ node: 'http://localhost:9200' })
+
+client.on('request', (err, event) => {
+  const { id } = event.meta.request
+  const { context } = event.meta
+  if (err) {
+    console.log({ error: err, reqId: id, context })
```
tylersmalley: I wonder if providing the winter key would make this example more clear, though they would obviously want to protect that it exists on the context before accessing like I did:

```
client.on('request', (err, event) => {
  const { id } = event.meta.request
  const { winter } = event.meta.context
  if (err) {
    console.log({ error: err, reqId: id, winter })
  }
})
```

lib/Transport.js
```diff
@@ -414,5 +424,12 @@ function randomSelector (connections) {
   return connections[index]
 }
 
+function generateRequestId () {
+  var maxInt = 2147483647
+  var nextReqId = 0
+  return function genReqId () {
+    return (nextReqId = (nextReqId + 1) & maxInt)
```
tylersmalley: 👌

docs/observability.asciidoc
delvedor: Good point!

delvedor:
retest

delvedor:
retest

Conky5:
Jenkins retest this please

docs/observability.asciidoc
```diff
@@ -0,0 +1,251 @@
...
+[source,js]
+----
+const { Client } = require('@elastic/elasticsearch')
+const client = new Client({
+  node: 'http://localhost:9200',
+  name: 'parent-client' // default to 'elasticsearch-js'
+})
+
+const child = client.child({
+  name: 'child-client'
```
spalger: maybe we should default this to `${parent.name} child`?

docs/usage.asciidoc
```diff
@@ -164,6 +164,14 @@ _Default:_ `null`
 |`querystring`
 |`object` - Custom querystring for the request. +
 _Default:_ `null`
+
+|`id`
+|`any` - Custom request id. _(overrides the top level request id generator)_ +
+_Default:_ `null`
+
+|`context`
+|`any` - Custom object per request. _(you can use it to pass some data to the clients events)_ +
```
spalger: Is this passed as is and never modified/cloned? I'm trying to make sure that there will be some way for people to establish a cache in a `WeakMap` for each request.

docs/configuration.asciidoc
```diff
@@ -133,6 +133,22 @@ function nodeSelector (connections) {
   return connections[index]
 }
 ----
+
+|`generateRequestId`
+a|`function` - function for generate the request id of every request. +
```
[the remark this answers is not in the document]
delvedor: > By default we call the internal generateRequestId() function to create another function with a closure scope. We don't do this for the generateRequestId() passed in options, do we want to?

Nope, I expect that the user passes me a function to generates unique ids, if this function requires a closure, the user will take care of that.

docs/configuration.asciidoc
```diff
@@ -133,6 +133,22 @@ function nodeSelector (connections) {
...
+[source,js]
+----
+function generateRequestId () {
+  var id = 0
+  return () => id++
+}
+----
+
+|`name`
+|`string` - The name to identify the client istance in the events. +
```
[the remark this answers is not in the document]
delvedor: It is passed around in all events and metadata, it is documented [here](https://github.com/elastic/elasticsearch-js/blob/correlation-id/docs/observability.asciidoc#client-name) :)

docs/observability.asciidoc
```diff
@@ -0,0 +1,251 @@
...
+    logger.info(req)
+  }
+})
+----
+
+The client emits the following events:
+[cols=2*]
+|===
+|`request`
+a|Emitted before to send the actual request to Elasticsearch.
```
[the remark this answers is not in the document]
delvedor: Yes, I'll add a note about this :)

docs/observability.asciidoc
```diff
@@ -0,0 +1,251 @@
...
+[source,ts]
+----
+  body: any;
+  statusCode: number | null;
+  headers: anyObject | null;
+  warnings: string[] | null;
+  meta: {
+    context: any;
+    name: string;
+    request: {
```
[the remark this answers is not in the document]
delvedor: Agree that `req` might not be a good name, I'll use `result`, since is the same used by the API methods.

> what about calling these requestParams, requestOptions, and requestId?

I prefer to have a nested object, since it is more future proof and I can easily extend it in the future.

docs/observability.asciidoc
delvedor: Initially, I thought about having an automatic name for every child client, such as `elasticsearch-js-child-${count}`. But other than increasing the internal complexity *(you will need a singleton to make sure to avoid duplicates)*, I think that if a user is interested in this feature, they will give a meaningful name.

docs/usage.asciidoc
delvedor: It is never cloned. I try to avoid cloning as much as possible since it will lead to bad performances.

README.md
```diff
@@ -143,7 +147,9 @@ The supported *request specific options* are:
   asStream: boolean, // default `false`
   compression: string, // default `false`
   headers: object, // default `null`
-  querystring: object // default `null`
+  querystring: object // default `null`,
+  context: object // default `null`,
+  id: any // default incr. integer
```
spalger: I think it would make sense to use the `unknown` type here, but it's only available in TS 3.0+, which is about a year old...

docs/configuration.asciidoc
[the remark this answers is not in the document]
spalger: What do you think about
> I think it would be really useful to pass this function the parameters received by the client so that they can be used to derive the request id, but maybe that's a bad idea as it encourages passing undocumented parameters...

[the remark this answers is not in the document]
delvedor: Sorry! Forgot about this one.
Can you make a code example?

README.md
delvedor: I didn't know about `unknown`, but I'm not sure it would make sense here.
I think that `any` is the most appropriate since I'm not doing anything with that but passing it around. The user can choose literally any value :)

docs/configuration.asciidoc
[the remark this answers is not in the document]
delvedor: Now I got you, yes we can do that :)

README.md
spalger: yeah, `unknown` has the same effect, but where it's different is when you/someone tries to use it internally. `any` will allow any operation such as `meta.request.id.toString()` works on `any`, but with `unknown` the type check will fail because there wasn't a check to make sure that `id` has a `toString` function.

As I type this though I realize that only makes a difference if the internals are implemented in TypeScript, which they aren't, so it makes sense to go for the most widely supported option (`any`).

docs/configuration.asciidoc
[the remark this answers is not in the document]
spalger:
```
function genReqId (requestParams, requestOptions) {
 ...
}
```

delvedor:
Jenkins retest this please

PR #845, opened 2019-05-10
Fix 843

### LINKED ISSUE #843: https connection doesn't work behind a socks5 proxy
## 🐛 Bug Report

The feature added to replace createNodeAgent doesnt work as expected 
https://github.com/elastic/elasticsearch-js/pull/810

When proxying the requests over socks5 through an ssh tunnel the library fails to find the host

After some debugging I found where the issue is but I don't have enough understanding of the library to apply it.

On the connection pool `urlToHost`  new URL will transform https://myelastic:443 and remove the port, since it is the https protocol

When running the request the host is:
```
host:
      [ 'Host',
        'myelastic:80' ] } }
```
If I do the following the request is done successfully:

```
const requestParams = this.buildRequestObject(params)
requestParams.port = undefined
```

If `requestParams.port = ''` , https.request host port will default to 80

## To Reproduce
Version 6.X
```
const SocksProxyAgent = require('socks-proxy-agent')

const client = new Client({
  node: 'https://myelasticsearch:443',
  agent: () => new SocksProxyAgent('socks5://127.0.0.1:7766')
})

client.search().then(response => {
  console.log(response)
})
```

## Expected behavior

Version 16.X
```
const { Client } = require('elasticsearch')
const SocksProxyAgent = require('socks-proxy-agent')

const client = new Client({
  host: 'https://myelasticsearch:443',
  createNodeAgent: () => new SocksProxyAgent('socks5://127.0.0.1:7766')
});

client.search().then(response => {
  console.log(response)
})
```

or the same as in the reproduce with:
```
requestParams.port = undefined
```

## Your Environment

- *node version*: v10.14.0
- `@elastic/elasticsearch` *version*: >=6.4.3
- *os*: Mac

delvedor:
Fixes #843.
See the issue for discussion.

@hensansi can you verify that this pr fixes your issue?

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -191,7 +191,8 @@ class Connection {
       path: '',
       href: url.href,
       origin: url.origin,
-      port: url.port,
+      // https://github.com/elastic/elasticsearch-js/issues/843
+      port: url.port !== '' ? url.port : undefined,
       headers: this.headers,
       auth: !!url.username === true || !!url.password === true
         ? `${url.username}:${url.password}`
--- a/test/unit/connection.test.js
+++ b/test/unit/connection.test.js
@@ -763,3 +763,34 @@ test('Util.inspect Connection class should hide agent and ssl', t => {
   roles: { master: true, data: true, ingest: true, ml: false } }`)
   )
 })
+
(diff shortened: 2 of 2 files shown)


hensansi:
It works for me now.

PR #852, opened 2019-05-16
Remove duplicated 'type' parameter from Bulk API

KnupMan:
There is already `type` parameter, so just removed the duplicated one.

image


KnupMan:
Created a new one, since commit was not signed.
https://github.com/elastic/elasticsearch-js/pull/853

PR #853, opened 2019-05-16
Remove duplicated 'type' parameter from Bulk API

KnupMan:
There is already `type` parameter, so just removed the duplicated one.

image


KnupMan:
@delvedor Okay, I understood. Thanks for the quick response!

delvedor:

PR #903, opened 2019-07-12
[DOCS] Fix broken links for ES API docs move

jrodewig:
elastic/elasticsearch/pull/44238 moved several Elasticsearch APIs to a REST APIs section.

As a result of that move, some pages were converted to anchored sections. This will fix links to those pages once the above PR is re-applied.


jrodewig:
Will do. Thanks @delvedor

PR #911, opened 2019-07-16
Added job configuration for Jenkins pipelines

delvedor:
This is just the job configuration, the `Jenkinsfile` will follow.


.ci/jobs/elastic+elasticsearch-js+jenkins-pipelines.yml
```diff
@@ -0,0 +1,51 @@
+---
+- job:
+    project-type: multibranch
```
Crazybus: All looks good except for some required fields like the job name. Basically the ones defined in here: https://github.com/elastic/apm-agent-go/blob/master/.ci/jobs/apm-agent-go-mbp.yml.

PR #924, opened 2019-07-26
[wip] Move from docker agent to labels

v1v:


Jenkinsfile
```diff
@@ -1,33 +1,58 @@
 pipeline {
-  agent {
-    docker {
-      image 'node:10-alpine'
-    }
+  agent { 
+    label 'linux'
```
delvedor: Why not `linux && immutable`?

Jenkinsfile
```diff
@@ -1,33 +1,58 @@
...
+        }
       }
     }
 
     stage('License check') {
       steps {
-        sh 'npm run license-checker'
+        script {
+          docker.image('node:10-alpine').inside(){
+            sh 'npm run license-checker'
```
delvedor: Why in the other steps you have added the `withEnv` block while here is not present?

Jenkinsfile
v1v: It can be as specific as you wish. I used linux just for a quick PR, but sure you can add `linux && immutable `, let me do it now

```suggestion
    label 'linux && immutable'
```

Jenkinsfile
v1v: That's a fair point, I created this PR as a draft example of how to use ephemeral workers with docker. It happened the npm cache folder was failing and I found it was required to use the `withEnv`, I iterated a few times, and the one for the `npm run license-checker` was not required though, actually I have not spent much longer looking for the main reason though.

I got a bit skeptical about using `node.image`, I'd rather prefer using `nvm` to specify what's the environment required rather than using docker within the Jenkinsfile. The main reasons:
- easy to debug locally as the Jenkinsfile should be an orchestrator,
- running docker.image step might imply some other deeper tweaks like the one I did with the `withEnv`.

delvedor: Thank you for the insight!
For now, I'll continue to use docker, since it's one of the most common ways to use Node.js in production, while `nvm` is more used for development.
We'll see how it goes! :)

PR #930, opened 2019-08-02
Fixed duplicate option in scroll method

hamdibayhan:


delvedor:

PR #935, opened 2019-08-08
Remove duplicated parameter from doc

vvision:
Hello !

I was reading [indices.clearCache](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/api-reference.html#_indices_clearcache) doc a few minutes ago and it seems to me, that the parameter named *index* appears twice.
So here is a PR.


vvision:
This make sense.
Thanks for mentioning the issue.

Issue #944, opened 2019-08-16
Updated the code generation script to use the new spec format


delvedor:
From Elasticsearch `v7.4` the new spec format will be used, the code generation script should be updated accordingly.

Related: https://github.com/elastic/elasticsearch/pull/42346

PR #945, opened 2019-08-16
Updated API reference doc

delvedor:
With this change, it should be more clear what to expect from this document and how to use the API.

cc @stacey-gammon @lukasolson

*Note: this pr cannot be merged before #944.*


stacey-gammon:
Awesome, thanks, this helps!

Issue #940, opened 2019-08-13
TypeError: Cannot read property 'resurrectTimeout' of undefined at ConnectionPool.resurrect


kokozhang:
## 🐛 Bug Report
Hello. When I called esClint.index() I got: 
`this.connections.get(this.dead[0])` is undefined at ConnectionPool.js

## To Reproduce

```js
const elasticsearch = require('@elastic/elasticsearch');

const esClient = new elasticsearch.Client({ node: ESHost.split(',').map(host => host.startsWith('http://') ? host : `http://${host}`), sniffOnStart: true });

esClient.index(q)
```

## Expected behavior
Index this document. And Code compiles without error

## Actual behavior
```js
TypeError: Cannot read property 'resurrectTimeout' of undefined
    at ConnectionPool.resurrect (/var/task/node_modules/@elastic/elasticsearch/lib/ConnectionPool.js:124:47)
    at Transport.getConnection (/var/task/node_modules/@elastic/elasticsearch/lib/Transport.js:332:25)
    at makeRequest (/var/task/node_modules/@elastic/elasticsearch/lib/Transport.js:124:30)
    at Transport.request (/var/task/node_modules/@elastic/elasticsearch/lib/Transport.js:316:15)
    at Promise (/var/task/node_modules/@elastic/elasticsearch/lib/Transport.js:90:14)
    at new WrappedPromise (/var/task/node_modules/async-listener/es6-wrapped-promise.js:13:18)
    at Transport.request (/var/task/node_modules/@elastic/elasticsearch/lib/Transport.js:89:14)
    at _index (/var/task/node_modules/@elastic/elasticsearch/api/api/index.js:148:12)
    at Client._lazyLoad [as index] 
```

## Your Environment

- *node version*: 8
- `@elastic/elasticsearch` *version*: 6.8.1
- *os*: Linux
- *Deployed in Lambda*

delvedor:
Hello!
I'm not able to reproduce the issue on my local machine, can you that with https://github.com/delvedor/es-reproduce-issue?

ganxiyun:
How to reproduce

1. ESHost list is provided in initialization
2. sniffOnStart is true
3. one connection returns error response so the connection is marked as dead before the finish of `update (connections)`

delvedor:
Are you able to reproduce it also with the latest release of the client?
[...]

PR #946, opened 2019-08-16
markAlive of connection correctly

ganxiyun:
See https://github.com/elastic/elasticsearch-js/issues/940


ganxiyun:
> Hello! Thank you for the fix! Can you also add a test that covers this case?

@delvedor changed the fix with test cases

ganxiyun:
@delvedor Can you please check the fix?

PR #967, opened 2019-09-13
Fix issue; node roles are defaulting to true when undefined (false) breaking usage of nodeFilter option

deuscapturus:
nodeFilter option was unable to filter nodes where master, data or ingest roles were false because they were defaulting to true when not set.

The example below demonstrates the problem:
```
// roles are returned from the elastic API as a list
roleslist = ['master', 'ingest']
console.log(roleslist)

// The client converts the list to a dictionary
const roles = roleslist.reduce((acc,role) => { acc[role] = true; return acc}, {})
console.log(roles)

// The client then assigns default values to undefined roles
console.log(
        Object.assign({'data': true, 'master': true, 'injest': true, 'ml': false}, roles)
)

```

Output:
```
[ 'master', 'ingest' ]
{ master: true, ingest: true }
{ data: true, master: true, injest: true, ml: false, ingest: true }
```
data is `true` when it should be `false`.

### DIFF
--- a/lib/pool/BaseConnectionPool.js
+++ b/lib/pool/BaseConnectionPool.js
@@ -212,9 +212,9 @@ class BaseConnectionPool {
         url: new URL(address),
         id: ids[i],
         roles: Object.assign({
-          [Connection.roles.MASTER]: true,
-          [Connection.roles.DATA]: true,
-          [Connection.roles.INGEST]: true,
+          [Connection.roles.MASTER]: false,
+          [Connection.roles.DATA]: false,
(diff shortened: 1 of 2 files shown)


deuscapturus:
tests added

deuscapturus:
oops.  I hijacked another test.  Let me try something different...

delvedor:
jenkins run the tests please

niksrc:
@delvedor any reason this change is not released for @elastic/elasticsearch@5 yet?

Issue #933, opened 2019-08-07
Define and export type SearchResponse


villasv:
## 🚀 Feature Proposal

Export the generic types defined in https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/typescript.html (specially the `SearchResponse<T>`)

## Motivation

This way people won't have to reimplement these types every time.

## Example

The exact same documentation example, but we would only have to define the `Source` type.

villasv:
I've seen the comment at https://github.com/elastic/elasticsearch-js/pull/813#discussion_r275340321 *(emphasis mine)*
> Because there is not an official spec from which I can generate the response types, and **it will make no sense export the types just for the search response**.

I can see that it would be too much work to give types to all possible response types, and it would be inconsistent to provide a response type just for the search.

Even so, I believe it's better to have this. Currently, the API is **already inconsistent** in that a few places are typed and others don't. I'd rather have *some* response types than none.

villasv:
My proposal: have a separate `d.ts` for types that are hand curated and warn about their incompleteness, so they can be exported like some kind of experimental types.

delvedor:
Hello!

> Currently, the API is already inconsistent in that a few places are typed and others don't. I'd rather have some response types than none.

Can you elaborate? What you mean by "inconsistent"?

> My proposal: have a separate d.ts for types that are hand curated and warn about their incompleteness, so they can be exported like some kind of experimental types.

This could be a viable solution, even if I don't like the idea of exporting only a subset of typed responses. I'll think about it.

lukeed:
Not having the API responses typed by default is pretty disappointing tbh – it's maybe half(?) the reason I wanted to use the definitions in the first place.

The documentation around using the library methods and/or constructing API requests is plentiful, but I have almost _zero_ idea what the response will look like until I actually run the query.

Would love to see this change in the future.

villasv:
> @delvedor Can you elaborate? What you mean by "inconsistent"?

I'm sorry, after re-reading I now understand what you meant with inconsistent. I agree this is not a good situation. Having a half-baked type system might as well worse than not having one at all and I imagine many people will get frustrated, just like now. OTOH I think we can grow to a reasonably comprehensive type system quite fast if we count on the community to come here fill the blanks whenever they find some. 

[...]

PR #970, opened 2019-09-18
Added Request/Response types definitions

delvedor:
With this pr we introduce the response type definitions for the following APIs:
* `Index`
* `Create`
* `Update`
* `Delete`
* `Search`
* `MSearch`
* `Bulk`

For now, we will not support more APIs because we are studying a more structured approach to be able to generate all the response definitions automatically.

Closes: https://github.com/elastic/elasticsearch-js/issues/933


api/ResponseTypes.ts
```diff
@@ -0,0 +1,122 @@
...
+export interface Search<T = any> {
+  took: number;
+  timed_out: boolean;
+  _scroll_id?: string;
+  _shards: Shards;
+  hits: {
+    total: {
+      value: number;
+      relation: string;
+    };
```
wylieconlon: This type is only correct when the server is running Elasticsearch 7.0 and above, and only when the request did not set the property `rest_total_hits_as_int`:

```
rest_total_hits_as_int
    (Optional, boolean) Indicates whether hits.total should be rendered as an integer or an object in the rest search response. Defaults to false.
```

delvedor: For now, this type definitions will land only for `7.x` and above, so it shouldn't be an issue.

> and only when the request did not set the property rest_total_hits_as_int:

How would you solve this? The easiest solutions that come to my mind is to use a union, in the same way I did in `BulkItems`.

wylieconlon: I think that would create extra work for the 80% case. If possible, I would suggest making the request object into a parameter on the type definition. I've made an example here for how I would write these types:

https://www.typescriptlang.org/play/index.html#code/JYOwLgpgTgZghgYwgAgEJwM4QMoTlBACwCUIMAHAexC2QG8AoZZ5OAczagjbjGGowB+AFzIAriADWISgHcQAbgYBfBgzABPcilz4ipCgIgAeUgEcxZMMgC84qTPkA+W8nOWM1iAA9IIACYY9MhcngD6YJRgcAA2YYTAYBhhmGGgYKJgUJbIysiCTCzoWLoEJGRUNCgAZPSFLCwJSaIgYgC2AEbQ9bnIosU4eGUGlbS1jA0NTRiidMgAbrGWLe1dUAohEDG8-CCiAORskPvIAD7I+xBm+6oNykoMCALWoaMQAHKr0KKl+hVGxjmoTAESisXiiWSqXSmWyKGULjsE0akNEAFYADQ9dicbg7ASzZRY+5qJ40F7-KoAeQ6ACsIAgMshfuVDFVAZtwpFonFpilkjDkPAYrQEa5kcxprMFksIOiMZttnxqAcricidiOFweMqaITiQ8yZ5OW8afTGQAmH5DP5s2hInpS4KLGLLZCYxX4vYXNW5LENHHar0zegakkMIA

api/ResponseTypes.ts
```diff
@@ -0,0 +1,122 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+export interface Search<T = any> {
```
wylieconlon: I would recommend `T = unknown` instead of `T = any`: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type

delvedor: I'm using `any` for consistency with the `RequestParams`, change that now would lead to a breaking change, unfortunately.

okachynskyy:
@delvedor Any updates on this PR?

Should I define `SearchResponse` / `Explanation` / `ShardsResponse` etc in my code like describe in documentation here?
https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/6.x/typescript.html

Or should I wait for adding these types directly in the Elastic package?

api/ResponseTypes.ts
Veetaha: Also, just for general awareness, I'd recommend looking at a [`discriminated union`](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions) feature.

Anyway, what looks simpler to me is just using two separate overloads:
```ts
// Don't export this one, because it is likely to be removed when the legacy interface is removed
interface BasicSearchResponse {
    /* all stable fields */
}

export interface SearchResponse extends BasicSearchResponse {
    hits: {
        total: {
            value: number,
            relation: "gte" | "eq" // @delvedor I suggest using an explicit string literal union here instead of just a `string`
        }
    }
}

// legacy, thus the interface name is verbose:
export interface SearchResponseWithTotalHintsAsInt extends BasicSearchResponse {
    hits: { total: number }
}

// don't export it for the same reason as BasicSearchResponse
interface BasicSearchParams { /* all fields except for rest_total_hits_as_int */ }

export interface SearchParams extends BasicSearchParams {
    rest_total_hits_as_int?: false;
}

export interface SearchParamsWithTotalHintsAsInt extends BasicSearchParams {
    rest_total_hits_as_int: true;
}

export class Client {
    search(params: SearchParams): Promise<SearchResponse>;
    // @deprecated: please use other overload
    search(params: SearchParamsWithTotalHintsAsInt): Promise<SearchResponseWithTotalHintsAsInt>;
}

```

api/ResponseTypes.d.ts
```diff
@@ -0,0 +1,132 @@
...
+  timed_out: boolean
+  _scroll_id?: string
+  _shards: Shards
+  hits: {
+    total: {
+      value: number
+      relation: string
+    }
+    max_score: number
+    hits: Array<{
```
Veetaha, line 16: @delvedor I'd suggest extract this inline object literal type into an interface. I in my code usually use not only the `Source` type, but the whole hit type because there is an id on that I'd like to use:
```ts
interface DocHit<TSource> {
    _id: string;
    _source: TSource;
}
```
If we don't extract this into an interface we won't be able to write the following code:

```ts
interface MyDocSource { foo: number }
// the type of hit should be DocHit<MyDocSource> but we don't have the DocHit type as per this PR right now
function processHit(hit: ??) {
    console.log("doc with id", hit._id, "has foo:", hit._source.foo);
}

const res = await client.search<MyDocSource>({ ... });

// notice that we cannot annotate the docHit param type with an interface name here to
res.body.hits.hits.forEach(docHit => processHit(docHit))
```

Veetaha:
@delvedor won't it be error-prone if the user can fully override the response type here?
```ts
client.msearch<MSearchResponse<Source>>(...);
```
The user can very easily mistakenly write an incorrect response type:
```ts
// notice the `M` prefix is mistakenly omitted
client.msearch<SearchResponse<Source>>(...);
```
My gut feeling is that we should pass not the response type as a whole, but only the generic parameters of that response type if there are any:
```ts
// Now this is way less verbose and less error-prone
client.msearch<Source>(...);
```

dgieselaar:
@delvedor looks good to me, this should allow us to tack our types on top of those of the client.

api/ResponseTypes.d.ts
```diff
@@ -0,0 +1,132 @@
...
+
+interface SearchResponse<TSource = unknown, TAggregations = unknown> {
+  took: number
+  timed_out: boolean
+  _scroll_id?: string
+  _shards: Shards
+  hits: {
+    total: {
+      value: number
+      relation: string
```
Veetaha, line 13:
```suggestion
      relation: "eq", "gte"
```

api/ResponseTypes.d.ts
```diff
@@ -0,0 +1,132 @@
...
+export {
+  SearchResponse,
+  MSearchResponse,
+  CreateResponse,
+  IndexResponse,
+  DeleteResponse,
+  UpdateResponse,
+  GetResponse,
+  BulkResponse
+}
```
Veetaha, line 132:
```suggestion
```
I'd suggest using `export interface  SearchResponse {}` right near interfaces declarations, this style of export is mostly avoided today

api/RequestTypes.d.ts
```diff
@@ -2464,21 +2464,106 @@ export interface WatcherPutWatch<T = RequestBody> extends Generic {
...
 }
 
-export interface XpackUsage extends Generic {
-  master_timeout?: string;
-}
+interface XpackUsageRequest extends GenericRequest {
+  master_timeout?: string;
+}
+
+export {
```
Veetaha, line 2486: What is the rationale for using this export style?

delvedor: As you can see in the `index.d.ts`, I'm exporting all the request/response type definitions with
```ts
export * from './api/RequestTypes'
export * from './api/ResponseTypes'
```
For some reason that I cannot fully understand, if  use the export style you are suggesting, I was able to also use interfaces that were not exported.
For example:
```ts
/* ResponseTypes.d.ts */
export interface Foo {}
interface Bar {}

/* index.d.ts */
export * from './api/ResponseTypes.d.ts'
export { Client }

/* code.ts */
// here I should be able to import Bar, but I can
import { Client, Foo, Bar } from './'
```

Then I discovered that using the "single export" statement solves this quirk.

api/ResponseTypes.d.ts
```diff
@@ -0,0 +1,132 @@
...
+
+interface UpdateResponse {
+  _shards: Shards
+  _index: string
+  _type: string
+  _id: string
+  _version: number
+  result: string
+}
```
Veetaha, line 82: Elasticsearch 7 supports returning the document source in update requests, e.g. the following request
```json
POST /<index>/_update/<id>
{
  "_source": "*",
  "doc": {
    "field": "val"
  }
}
```
gives the following response
```json
{
  "_index" : "<index>",
  "_type" : "_doc",
  "_id" : "<id>",
  "_version" : 5,
  "result" : "updated",
  "_shards" : {
    "total" : 2,
    "successful" : 1,
    "failed" : 0
  },
  "_seq_no" : 53,
  "_primary_term" : 1,
  "get" : {
    "_seq_no" : 53,
    "_primary_term" : 1,
    "found" : true,
    "_source" : {
      "field": "val"
    }
  }
}
```
So the user should be able to specify the type of that `_source` (furthermore this `get` fields should be an optional property in `UpdateResponse<T>`.


@delvedor also note that the shards object I got from Elasticsearch doesn't contain `skipped` fields, this is either a bug or an intended behavior, so if the latter we should mark it as optional: `interface Shards { skipped?: number }`

Also note that there are two more fields:
```json
{
  "_seq_no" : 53,
  "_primary_term" : 1
}
```
I think they also should be included in `interface UpdateResponse`

api/RequestTypes.d.ts
Veetaha: Okay, this is indeed a quick of `.d.ts.` files, [all their symbols are exported by default](https://github.com/Microsoft/TypeScript/issues/19545). However, I think there should not be any private interfaces in `.d.ts` files, users should be able to access any interface that the can potentially use in their code, e.g. why not exporting `interface Shards`?
The user might want to write the following code:
```ts
const res = await client.update(...);
const shards: Shards = res._shards;
processShards(shards);

// Why not giving the possibility to use the Shards type?
function processShards(shards: Shards) {}
```

delvedor:
> @delvedor won't it be error-prone if the user can fully override the response type here?

@Veetaha it's something that could happen, yes. The main issue right now is that we don't have all the response types, and while we are working on it, this is not something that will change soon.
I fear that if some APIs offer you a way to define the full body response, others allow you to specify a part of it, and others do not provide anything at all, what you will get at the end is a massive source of confusion. Using the same approach for every API until we have a stable enough type definition provides a predictable API, that is easy to think about and do not require you to go and read the documentation/type definitions every time you are writing a line of code.

api/RequestTypes.d.ts
delvedor: As you know, the API surface of Elasticsearch is massive, and given that we are doing this work manually, it's very easy to miss something.
There might be a `ShardsResponse` that is different from `Shards` (I don't think this is the case, but it might happen), and I would love to avoid doing too many breaking changes because I wasn't cautious about what I'm exposing.
Adding these response types will be a very long work that will be done during time, and a big part of it will be fix and improve the types already written.
Thank you for your suggestions!

api/ResponseTypes.d.ts
restrry: +1 here. I hit this problem a couple of days ago. An example from Kibana codebase: 
https://github.com/elastic/kibana/blob/master/x-pack/plugins/logstash/server/types.ts#L11-L12

scripts/utils/generateRequestTypes.js
```diff
@@ -127,6 +133,18 @@ export interface ${toPascalCase(name)}${body ? `<T = ${bodyGeneric}>` : ''} exte
         return type
     }
   }
+
+  function generateExport (spec) {
+    const api = Object.keys(spec)[0]
+    const name = api
+      .replace(/\.([a-z])/g, k => k[1].toUpperCase())
```
restrry, line 140: We repeat the same operation across several `generate` utils. I'm wondering if we can normalize data once in a centralized manner.

api/ResponseTypes.d.ts
restrry: I'm surprised that we are able to generate request types from the API spec, but no responses. https://github.com/elastic/elasticsearch/blob/d7bbc9df1d3edf47f8d1c2f0983c88c64d78a4b4/rest-api-spec/src/main/resources/rest-api-spec/api/search.json#L231 
@delvedor Do you know if the elasticesearch team is going to introduce a spec for API responses as well?

scripts/utils/generateMain.js
```diff
@@ -181,36 +192,37 @@ function buildMethodDefinition (api, name, hasBody) {
   const Name = toPascalCase(name)
   const bodyType = ndjsonApiKey.includes(Name) ? 'RequestNDBody' : 'RequestBody'
   const defaultBodyType = ndjsonApiKey.includes(Name) ? 'Record<string, any>[]' : 'Record<string, any>'
+  const responseType = supportedResponses.includes(Name) ? `Res.${Name}Response` : 'Record<string, any>'
```
restrry, line 195: optional: why not export responses used `Response` namespace?

api/ResponseTypes.d.ts
```diff
@@ -0,0 +1,132 @@
...
+  skipped: number
+}
+
+interface Explanation {
+  value: number
+  description: string
+  details: Explanation[]
+}
+
+interface MSearchResponse<TSource = unknown, TAggregations = unknown> {
```
restrry, line 47: don't see it's used in `index.d.ts` file

api/RequestTypes.d.ts
```diff
@@ -1311,7 +1311,7 @@ export interface Termvectors<T = RequestBody> extends Generic {
   body?: T;
 }
 
-export interface Update<T = RequestBody> extends Generic {
+interface UpdateRequest<T = RequestBody> extends GenericRequest {
```
restrry, line 1314: interestingly, we use `RequestBody` as a default value here, but `Record<string, any>` in `index.d.ts`.

villasv:
Just wanted to chime in once more and thank @delvedor for the hard work on this, once release I'm going to slap this into our codebase like it's Flex Tape.

api/ResponseTypes.d.ts
```diff
@@ -0,0 +1,132 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+interface SearchResponse<TSource = unknown, TAggregations = unknown> {
```
restrry, line 5: It's not meant to be used in `api/kibana.d.ts`?

delvedor:
Closed in favor of #1358.

PR #980, opened 2019-10-04
Docs: Return super in example Transport subclass

nathanmoon:
If called without a callback, the request method returns a Promise, so
when calling into super.request, the result should be returned to
maintain promise behavior.


elasticcla:
Hi @nathanmoon, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-js/pull/980.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

nathanmoon:
I added my email.  Do I need to do something to trigger the check to run again?

PR #983, opened 2019-10-07
updated usage doc to provide information on error messages

dhuang612:


dhuang612:
thanks for the feedback I'll get the changes made!

dhuang612:
okay, added in the information into the table.

PR #1028, opened 2019-12-18
Removed repeated `type` field in `bulk` api docs 6.x

lmammino:
The `type` field is repeated twice in the `bulk` API reference documentation:

Screenshot 2019-12-18 at 15 48 42


lmammino:
Thanks for your detailed explanation @delvedor :) All the best 👋

PR #1053, opened 2020-01-15
Change TransportRequestOptions.ignore to number[]

joshdifabio:
The underlying JavaScript handles arrays of any length. The current type of TransportRequestOptions.ignore, which is [number], only allows arrays with length==1, which looks like a mistake, and prevents the ignore API from being usable when we want to ignore multiple error statuses (which I do).


delvedor:
jenkins run the tests please

joshdifabio:
> Can you sign the CLA? :)

Done.

PR #1076, opened 2020-01-31
Add examples to reference

delvedor:
With this pr we will include all the examples present in `docs/examples` to their corresponding API in the reference doc.


docs/reference.asciidoc
```diff
@@ -8575,6 +8584,7 @@ client.sql.query({
 })
 ----
 link:Execute SQL[Reference]
+{jsclient}/sql.query_examples.html[Example]
```
szabosteve: I think this is the link that causes the problem.

```suggestion
{jsclient}/sql_query_examples.html[Example]
```

PR #1092, opened 2020-02-22
typo

avetisk:

### DIFF
--- a/docs/reference.asciidoc
+++ b/docs/reference.asciidoc
@@ -21,7 +21,7 @@ const result = await client.search({
   maxRetries: 3
 })
 
-// calback API
+// callback API
 client.search({
   index: 'my-index',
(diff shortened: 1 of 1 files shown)


avetisk:
> Hello! Thank you for contributing!
> Could you sign the CLA? :)

I did it, even twice I think, but the status is not updated here.

delvedor:
@avetisk now it worked, sorry for the noise!
Since `reference.asciidoc` is generated, your change will be overridden in the future.
I would kindly ask you to close this pr and open a new one against master and then fix the script that generates this file, which is:
https://github.com/elastic/elasticsearch-js/blob/94d4ac752576947a9228542bf6b958bb1d3a1907/scripts/utils/generateDocs.js#L73

avetisk:
#1109

Issue #227, opened 2015-06-13
bulk api helper


spinscale:
There is a nice package called [elastic-queue](https://www.npmjs.com/package/elastic-queue/) that allows you to add items to a queue and that one is batched based on size or age, similar to the java [BulkProcessor](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/action/bulk/BulkProcessor.java)

Can we get that helper as part of the official package, or at least an official helper package? I saw that you did something like that in `makelogs` already, so I think it would be a nice addition.

Please fix the subject to something more JS'y :-)

spalger:
Haha :) I think a set of helpers for esjs would be a great thing. Finding the time to do it is tough these days.

spinscale:
I still think that this is a valid feature request for the core client as it is still a missing feature. The python client has a similar feature for example. See https://elasticsearch-py.readthedocs.io/en/master/helpers.html

PR #1107, opened 2020-03-10
Client helpers

delvedor:
With this pr we add some useful and needed client helpers:

- Bulk
- Scroll search
- Scroll documents
- Search

We might add more helpers in the future, but initially, we will start with the ones described above.

Closes: https://github.com/elastic/elasticsearch-js/issues/227


delvedor:
Jenkins run docs build.

PR #1132, opened 2020-03-30
Switched request and response generics position

### LINKED ISSUE #1130: Switch Request and Response generics
In #1119 we introduced a big change in how the type definitions work, and I'm very happy with the result, the code is more idiomatic and it's easier to use the TypeScript features.
While working on #970, I've noticed that there is still something we can do to improve even further the developer experience.
By reading this issue tracker and around on the internet, people is more interested in the response types rather than the request types, which in most of the cases are good enough with the body set to `Record<string, any>`.
Currently, the generics allow you to specify the request body, the response body  and the context in this order:
```ts
client.search<RequestBody, ResponseBody, Context>(...)
```
I'm wondering if it does make more sense to have the `ResponseBody` generic as the first parameter instead since it will be used more often for defining Source objects and so on. While my OCD screams when seeing about this, I think it makes way more sense given that people will use the Response types more often.
```ts
client.search<ResponseBody, RequestBody, Context>(...)
```

What do you think?

cc @Mpdreamz @joshdover @restrry @nreese @dgieselaar @fox1t @villasv

delvedor:
As titled, closes #1130.


lib/errors.d.ts
```diff
@@ -2,42 +2,42 @@
...
+export declare class NoLivingConnectionsError<TResponse = Record<string, any>, TContext = unknown> extends ElasticsearchClientError {
   name: string;
   message: string;
-  meta: ApiResponse;
+  meta: ApiResponse<TResponse, TContext>;
   constructor(message: string, meta: ApiResponse);
 }
 
-export declare class SerializationError extends ElasticsearchClientError {
+export declare class SerializationError<TResponse = Record<string, any>, TContext = unknown> extends ElasticsearchClientError {
```
restrry, line 33: Is it referenced anywhere?

test/types/helpers.test-d.ts
```diff
@@ -129,14 +128,77 @@ expectError(
...
+    };
+    aggregations?: any;
+  }
+  
+  interface Source {
+    foo: string
+  }
+
+  async function test () {
+    const scrollSearch = client.helpers.scrollSearch<Source, SearchResponse<Source>>({
```
restrry, line 184: shouldn't it follow the same type order as other generics? `<ResponseBody, RequestBody, Context, Document>`?

test/types/api-response-body.test-d.ts
```diff
@@ -126,3 +126,36 @@ expectError(
   expectType<SearchResponse<Source>>(response.body)
   expectType<string>(response.meta.context)
 }
+
+// Send request body as string
+{
```
restrry, line 131: don't see any tests for callback form.

lib/errors.d.ts
delvedor: `E_TOO_MAY_COPY_PASTE` 😄

test/types/helpers.test-d.ts
delvedor: The reason why I put `TDocument` as the first parameter, is because the `scrollSearch` returns `AsyncIterable<ScrollSearchResponse<TDocument, TResponse, TContext>>`, and `ScrollSearchResponse` is defined as follows:
```ts
export interface ScrollSearchResponse<TDocument = unknown, TResponse = Record<string, any>, TContext = unknown> extends ApiResponse<TResponse, TContext> {
  clear: () => Promise<void>
  documents: TDocument[]
}
```
Having `TDocument` as first parameter would allow users to access directly `response.documents` instead of accessing the full search response.

delvedor:
> I do agree the current update brings more value and inlined with typical network client interfaces (axios)

👍 

> I concern about the update process. Would it be hard to release a major release to follow the proper semver?

I can't release a major until Elasticsearch 8, as the client strictly follows the Elastic Stack release train.
I'm fine doing a breaking change specifically for the typescript definitions for two reasons:
1. It will improve massively the developer experience.
2. The TypeScript core team does release breaking changes in minor releases, so I'm assuming the community is ok with this behavior.

PR #1133, opened 2020-03-30
Prevent upstream error of getting property 'username' or 'password' o…

matthewadams:
…f undefined

The prior implementation returns `undefined`, causing line 219 to fail, because it's getting property `username` of `undefined`.  This improves the situation by at least returning an empty object.

### DIFF
--- a/lib/ConfigurationHelpers.js
+++ b/lib/ConfigurationHelpers.js
@@ -1,0 +1,37 @@
+// Licensed to Elasticsearch B.V under one or more agreements.
+// Elasticsearch B.V licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information
+
+'use strict'
+
+const { ConfigurationError } = require('./errors')
+
+function getAuth (nodes) {
+  nodes = nodes || []
+  nodes = Array.isArray(nodes) ? nodes : [nodes]
+
+  for (const node of nodes) {
+    const auth = getUsernameAndPassword(node)
+    if (auth.username) return auth
+  }
+
+  return null
+
+  function getUsernameAndPassword (node) {
+    let it = node && (node.url || node)
+    try {
+      it = it instanceof URL ? it : new URL((it && it.toString()))
+    } catch (e) {
+      throw new ConfigurationError(`invalid url in option "node${node.url ? '.url' : ''}": ${e.message}`)
+    }
+
+    return {
+      username: decodeURIComponent(it.username),
+      password: decodeURIComponent(it.password)
+    }
+  }
+}
+
+module.exports = {
+  getAuth
+}
--- a/index.d.ts
+++ b/index.d.ts
@@ -55,7 +55,7 @@ interface ClientExtends {
 // /Extend API
 
 interface NodeOptions {
-  url: URL;
+  url: string | URL;
   id?: string;
   agent?: AgentOptions;
   ssl?: TlsConnectionOptions;
--- a/index.js
+++ b/index.js
@@ -243,6 +243,8 @@ function getAuth (node) {
         username: decodeURIComponent(node.url.username),
         password: decodeURIComponent(node.url.password)
       }
+    } else {
+      return {}
     }
   }
 }
--- a/index.js
+++ b/index.js
@@ -5,7 +5,6 @@
 'use strict'
 
 const { EventEmitter } = require('events')
-const { URL } = require('url')
 const debug = require('debug')('elasticsearch')
 const Transport = require('./lib/Transport')
 const Connection = require('./lib/Connection')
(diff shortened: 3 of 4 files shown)


matthewadams:
@delvedor I've already signed the CLA.  I'll see about a test case when I have time.  I'm quite busy ATM.  I've added you as a collaborator in my fork at https://github.com/matthewadams/elasticsearch-js; if you would like to create a test case, create a configuration that sets `node.url` of the options object given to the `Client` constructor to a _string_ URL instead of a `URL` object.  That's enough to  reproduce the error.

penge:
How about this?

1) Always return `{ username, password }`, not an empty object
2) Check if `node` is an object before accessing `node.url`

**Code:**

```js
function getUsernameAndPassword (node) {
  if (typeof node === 'string') {
    const { username, password } = new URL(node)
    return {
      username: decodeURIComponent(username),
      password: decodeURIComponent(password)
    }
  }

  if (typeof node === 'object' && node.url instanceof URL) { // 2
    return {
      username: decodeURIComponent(node.url.username),
      password: decodeURIComponent(node.url.password)
    }
  }

  return {
    username: undefined, // 1
    password: undefined
  }
}
```

**Output:**

```js
// string => { username: "bart", password: "123" }
console.log(
  getUsernameAndPassword('http://bart:123@localhost:9200')
);

// object => { username: "lisa", password: "456" }
console.log(
  getUsernameAndPassword({
    url: new URL('http://lisa:456@localhost:9200')  
  })
);

// something else => { username: undefined, password: undefined }
console.log(
  getUsernameAndPassword(42)
);
```

matthewadams:
done

penge:
@matthewadams

Can I add the test?

matthewadams:
I just pushed what I think it ought to be -- hope you like.

lib/ConfigurationHelpers.js
```diff
@@ -0,0 +1,37 @@
...
+    return {
+      username: decodeURIComponent(it.username),
+      password: decodeURIComponent(it.password)
+    }
+  }
+}
+
+module.exports = {
+  getAuth
+}
```
penge, line 37: Let's rename this to `getCredentials()` and simplify a bit:

```js
// credentials.js

const { ConfigurationError } = require('./errors')

/**
 * Returns { username, password }, or throws ConfigurationError if node invalid
 * 
 * node can be:
 * A) 'http://abc:123@localhost:9200'
 * B) { url: new URL('http://abc:123@localhost:9200') }
 *
 * See: elasticsearch-js/test/unit/client.test.js
 */
function getCredentials (node) {
  try {
    const url = (node.url instanceof URL) ? node.url : new URL(node)
    return {
      username: decodeURIComponent(url.username),
      password: decodeURIComponent(url.password)
    }
  } catch (e) {
    throw new ConfigurationError(`invalid url in node "${node}": ${e.message}`)
  }
}

module.exports = { getCredentials }
```

Now, we can return credentials for the specific node only and move any other complexity (decision) outside, to `index.js`:

```js
// index.js

// ...
const node = Array.isArray(opts.nodes) ? opts.nodes[0] : opts.node
const credentials = getCrendentials(node)
// ...

// See also: elasticsearch-js/test/unit/client.test.js
```

penge:
I agree. As being used at only one place, that is `index.js`, there is no really need to decouple it.

I also agree that one way to do it, would be better. If it's URL instance or string. I would probably vote for the string as it's an easier, easier to move around format. The documentation would be therefore easier too.

I think we can skip the check if the `node` is an `URL` or a `string` because that check would not be enough in case of a `string`. It can still throw a `TypeError` if the URL string is not of a valid format. We can instead "expect" the correct format to come in, and let the `URL()` constructor do all the work.

matthewadams:
I spun the code in question out of `index.js` to make it testable.  I see no issue with spinning it out because `ConfigurationHelper.js` is down in the guts of the project.  I don't think anyone will care that it's spun out.  I'd encourage you to continue to pursue modularity like that in the name of testability.  As you design something, you should be asking yourself, "How do I test this?"  If it turns out to be hard to test, that's a yellow flag.  If it's impossible to test, that's a red flag.  Since our job as software developers is not to deliver code, but code that is _known to work_, I stand behind my work.  If you are putting something together with multiple parts, you're in a better situation if you're putting together parts that are known to work.

Also, most configuration frameworks deal with strings, not `URL` objects.  If you want to be dogmatic, yes, `node.url` should _only_ be an `instanceof URL`, but practically speaking, I don't know why you didn't make the code accept `string` or `URL` in `node`.  That would've been a good usability improvement and simplified things.  My recommendation is to deprecate `node.url` and accept `string` or `URL` in `node` (and call it `url` instead of `node`, but that's another discussion).  Remember that source code is for the humans, not the computers.

@penge & @delvedor, I have invited y'all to collaborate on my fork of this repo via the repo's settings.  I hope you appreciate my contributions of both code and advice.  Please complete the work as you see fit.

penge:
Thank you for the lecture 😂

matthewadams:
@delvedor I kindly invite you to complete those changes.  I don't have time right now.

Issue #1108, opened 2020-03-10
Fix flaky test


delvedor:
The following test passes consistently in localhost, but it's flaky on CI. It should be refactored to avoid this issue. Very likely the problem is caused due to the use of timers.

https://github.com/elastic/elasticsearch-js/blob/94d4ac752576947a9228542bf6b958bb1d3a1907/test/behavior/sniff.test.js#L114-L149

penge:
Do you have the log from the CI?

Wasn't the reason for the fail `TimeoutError` by any chance?

delvedor:
This is the log for the failure: https://github.com/elastic/elasticsearch-js/runs/489895557.

You can reproduce it yourself by forking this repository and playing with GitHub actions :)

penge:
Cool! :) I'll check it out!

PR #1142, opened 2020-04-04
Fix flaky test

penge:
Fixes flaky **'Sniff interval'** in `sniff.test.js`.

The problem was caused by not waiting for the asynchronous `kill()` to finish the task,
before making a request about the current `connectionPool.size`.

The error was:

```js
t.strictEqual(client.connectionPool.size, 3)

// expected: 3  => because one node should be killed
//   actual: 4  => because the request was made before the node was killed
```

The solution was to update `kill()` by adding a `callback` and calling it after the `stop()` in `stoppable.js` has finished. This was then used in `sniff.test.js` and `resurrect.test.js`.

Logging in `shutdown()` is also improved.

Unused `spawn()` is removed.

Closes https://github.com/elastic/elasticsearch-js/issues/1108.

### DIFF
--- a/test/behavior/sniff.test.js
+++ b/test/behavior/sniff.test.js
@@ -111,8 +111,8 @@ test('Should handle hostnames in publish_address', t => {
   })
 })
 
-test('Sniff interval', { skip: 'Flaky on CI' }, t => {
-  t.plan(10)
+test('Sniff interval', t => {
+  t.plan(13)
 
   buildCluster(({ nodes, shutdown, kill }) => {
     const client = new Client({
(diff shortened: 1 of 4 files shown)


penge:
Here it is @delvedor ;)

penge:
I've added some more `setTimeout` calls to test that the **SNIFF** event is only run when:

```
1) delay is greater than sniffInterval
2) delay is greater than time of last sniff + sniffInterval
```

This is also mentioned in the added comments.

I think it's now good to merge.

test/behavior/sniff.test.js
```diff
@@ -111,39 +111,44 @@ test('Should handle hostnames in publish_address', t => {
...
-      client.info(t.error)
+      kill('node1', () => {
+        client.info(t.error) // meets 1) and 2)
+      })
     }, 150)
 
-    setTimeout(() => {
-      t.strictEqual(client.connectionPool.size, 3)
-    }, 200)
-
```
delvedor, line 146: Great catch! I would kindly ask you to revert this change, leave the test as it was before and keep just the `kill` fix :)

penge: Thanks! Can you please explain to me why to revert the test?

I see 4 reasons why the new test is better. The previous test would _NOT_ work (1), would not be complete (2), can randomly fail (3), and has an unrelated assertion (4).

Here's the detail:

1) By not waiting for `kill()` to finish, `shutdown()` would try to kill 4 nodes instead of 3, because of outdated `nodes`. This would result in the following error:

```
kill('node1')
> kill node1

shutdown()
> kill node0
> kill node1  // Should NOT be here! Cannot kill!
> kill node2
> kill node3

Error [ERR_SERVER_NOT_RUNNING]: Server is not running.
```

2) The previous test does not test **SNIFF** event earlier than `sniffInterval` and earlier than "time of last sniff" + `sniffInterval`.

3) The previous statement:

```js
setTimeout(() => {
  t.strictEqual(client.connectionPool.size, 3)
}, 200)
```

Has an expectation that depends on **SNIFF** to run `this.connectionPool.update(hosts)`. Also depends on lazy loading of `info`, connection speed, time of killing the node.

4) Similar to point 3). This test is primarily about **'Sniff interval'** and not about nodes. Also, `client.on(events.SNIFF, (err, result) => { ... }` already has the assertion needed, and compares 4 nodes with 4, on the first run, and 3 with 3, on the second run.

Please explain. I don't see the reasons to revert the test.

penge:
No problem. Cool

Issue #1154, opened 2020-04-11
Failing "Should not retry if the body is a stream"


penge:
## 🐛 Bug Report

Failing test found. Only on CI.

Logs: https://github.com/elastic/elasticsearch-js/pull/1142/checks?check_run_id=579324431

```
# Subtest: Should not retry if the body is a stream

found: 0
wanted: 1
compare: ===
at:
    line: 667
    column: 9
    file: test/unit/transport.test.js
stack: |
    test/unit/transport.test.js:667:9
    onResponse (lib/Transport.js:157:16)
    ClientRequest.<anonymous> (lib/Connection.js:94:9)
```

delvedor:
I've opened https://github.com/elastic/elasticsearch-js/pull/1158 with the fix for this issue as well.
Thank you for reporting!

PR #1158, opened 2020-04-14
Fix flaky test

delvedor:
As titled.
The root cause of the issue was that we were testing with timers, and timers are not reliable between different platforms because they can cause timing issues like in this test. I've refactored the code to not depend on timers, but mock the `Date` constructor instead and use the internal state of the client to figure out if a sniff operation is still running or not.

A big thanks to @penge for his initial research.

Closes: #1108 #1142 #1154

### DIFF
--- a/test/behavior/sniff.test.js
+++ b/test/behavior/sniff.test.js
@@ -6,6 +6,8 @@
 
 const { test } = require('tap')
 const { URL } = require('url')
+const lolex = require('lolex')
+const workq = require('workq')
 const { buildCluster } = require('../utils')
 const { Client, Connection, Transport, events, errors } = require('../../index')
 
@@ -111,8 +113,10 @@ test('Should handle hostnames in publish_address', t => {
   })
 })
 
-test('Sniff interval', { skip: 'Flaky on CI' }, t => {
-  t.plan(10)
+test('Sniff interval', t => {
+  t.plan(11)
+  const clock = lolex.install({ toFake: ['Date'] })
+  const q = workq()
 
   buildCluster(({ nodes, shutdown, kill }) => {
     const client = new Client({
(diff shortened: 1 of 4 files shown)


delvedor:
I've fixed #1154 as well. Also here, the issues might have been caused by timing issues. Now the fake server returns a 504 error, which will trigger a retry internally.

test/behavior/sniff.test.js
```diff
@@ -132,19 +136,47 @@ test('Sniff interval', { skip: 'Flaky on CI' }, t => {
...
+    // at the right moment, we verify that the transport
+    // is no longer sniffing
+    function waitSniffEnd (callback) {
+      if (client.transport._isSniffing) {
+        setTimeout(waitSniffEnd, 500, callback)
+      } else {
+        callback()
+      }
+    }
   })
```
penge, line 180: Didn't know about `client.transport._isSniffing` Seems like a nice trick. I was dealing with the same, so what I did, I just ignored the last sniff. Because even when expecting it to happen 3-times (2 nodes killed) or 2-times (1 node killed), it had sometimes a chance to be called one more time.
500 seems like enough of waiting time

test/utils/buildCluster.js
```diff
@@ -56,14 +56,17 @@ function buildCluster (options, callback) {
...
   }
 
-  function kill (id) {
+  function kill (id, callback) {
     debug(`Shutting down cluster node '${id}' (cluster id: '${clusterId}')`)
-    nodes[id].server.stop()
+    const node = nodes[id]
     delete nodes[id]
     delete sniffResult.nodes[id]
+    node.server.stop(callback)
```
penge, line 69: I realized later that sniff result endpoint is mocked for the tests. So if I waited or didn't wait, I think I was getting the same results. But I think in case of resurrect it is needed to wait actually.

test/behavior/sniff.test.js
delvedor: > 500 seems like enough of waiting time

500 is just a random number, it could also be 50, the point here is that I'm not waiting for 500 and then doing something, but I'm checking a value that represents what I need every 500ms before starting the next operation. In this way the test code is more reliable :)

penge: I got lost at random number :)

Issue #1153, opened 2020-04-11
Cannot read property 'resurrectTimeout' of undefined


penge:
## 🐛 Bug Report

Happened on CI only. The message was `Cannot read property 'resurrectTimeout' of undefined`

See logs:
- https://github.com/elastic/elasticsearch-js/pull/1142/checks?check_run_id=578970718
- https://github.com/elastic/elasticsearch-js/pull/1142/checks?check_run_id=578970732

Related to https://github.com/elastic/elasticsearch-js/issues/940

```
Cannot read property 'resurrectTimeout' of undefined

stack:
  ConnectionPool.resurrect (lib/pool/ConnectionPool.js:101:47)
  ConnectionPool.getConnection (lib/pool/ConnectionPool.js:167:10)
  Transport.getConnection (lib/Transport.js:348:32)
  makeRequest (lib/Transport.js:123:30)
  onResponse (lib/Transport.js:150:13)
  ClientRequest.request.on.err (lib/Connection.js:103:9)

at:
  line: 101
  column: 47
  file: lib/pool/ConnectionPool.js
  function: ConnectionPool.resurrect
```

Looks like there can be `dead` with an ID that hasn't the mathing `connection`. And therefore, `connection.resurrectTimeout` fails resulting in `Cannot read property 'resurrectTimeout' of undefined`.

penge:
Here's a detailed run where we can see what is happening:

```
2020-04-11T17:35:42.2635027Z     markDead-dead [ 'node1' ]
2020-04-11T17:35:42.2635481Z     markDead-connection [ 'node0', 'node3' ]
2020-04-11T17:35:42.2635715Z     resurrect-dead [ 'node1' ]
2020-04-11T17:35:42.2635956Z     resurrect-connection [ 'node0', 'node3' ]
2020-04-11T17:35:42.2636166Z     resurrect-dead [ 'node1' ]
2020-04-11T17:35:42.2636380Z     resurrect-connection [ 'node0', 'node3' ]
2020-04-11T17:35:42.2636612Z     markDead-dead [ 'node1', 'node2' ]
2020-04-11T17:35:42.2636968Z     markDead-connection [ 'node0', 'node3' ]
2020-04-11T17:35:42.2637233Z     not ok 7 - Cannot read property 'resurrectTimeout' of undefined
```

More here: https://github.com/elastic/elasticsearch-js/pull/1142/checks?check_run_id=579393215

Why this is happening:

[...]

Issue #1122, opened 2020-03-22
Cannot read property 'resurrectTimeout' of undefined


kulakowka:
Greetings! 

I found an error in the code:
 
https://github.com/elastic/elasticsearch-js/blob/1b638d69925d64a7402617265e739dc85b316ca1/lib/pool/ConnectionPool.js#L107

```js
    const connection = this.connections.find(c => c.id === this.dead[0])

    if ((opts.now || Date.now()) < connection.resurrectTimeout) {
      debug('Nothing to resurrect')
      callback(null, null)
      return
    }
```

If the connection is not found. Then the code will throw an error further: **Cannot read property 'resurrectTimeout' of undefined**

I just got a lot of such errors in production. 

<img width="922" alt="Screen Shot 2020-03-22 at 8 57 21 AM" src="https://user-images.githubusercontent.com/557190/77243449-2a6e2600-6c1b-11ea-91ec-e3b78391ceb0.png">

delvedor:
Hello! Thank you for reporting this!
It looks like the issue you have encountered is hiding another bug.
That code should always have a "valid" dead connection because few lines above, we are checking that the `dead` array contains something.
So very likely, the `dead` array is not cleaned up properly.

Which version of the client and Elasticsearch are you using?
Where are you running Elasticsearch?
Can you paste here the options you are using for configuring the client?

Thanks!

kulakowka:
Client config: 
```yml
elastic:
  client:
    maxRetries: 6
    requestTimeout: 10000
    sniffOnConnectionFault: true
    sniffOnStart: true
    nodes:
      - http://***.***.**.**:9201
      - http://***.***.**.**:9201
      - http://***.***.**.**:9201
    agent:
      keepAlive: false
      keepAliveMsecs: 1000
      maxSockets: 256
      maxFreeSockets: 256
```

Client version:

```json
{
  "@elastic/elasticsearch": "^7.6.1"
}
```

Elastic version:

```json
{
  "name" : "mvs-elastic-01-mvs",
  "cluster_name" : "mvs",
  "cluster_uuid" : "-q1EWefBQa2auCjGwMLnfw",
  "version" : {
    "number" : "7.4.2",
    "build_flavor" : "default",
    "build_type" : "deb",
    "build_hash" : "2f90bbf7b93631e52bafb59b3b049cb44ec25e96",
    "build_date" : "2019-10-28T20:40:44.881551Z",
    "build_snapshot" : false,
[...]

PR #1159, opened 2020-04-14
Fix 1153

delvedor:
It might happen that `markDead` is called just after a pool update, and in such case, we will add to the dead list a node that no longer exists.
This pr fixes this issue :)
A big thanks to @penge for his initial research.

Closes: #1153 
Related: https://github.com/elastic/elasticsearch-js/issues/1122 https://github.com/elastic/elasticsearch-js/issues/1153

### DIFF
--- a/lib/pool/ConnectionPool.js
+++ b/lib/pool/ConnectionPool.js
@@ -60,7 +60,17 @@ class ConnectionPool extends BaseConnectionPool {
     const { id } = connection
     debug(`Marking as 'dead' connection '${id}'`)
     if (this.dead.indexOf(id) === -1) {
-      this.dead.push(id)
+      // It might happen that `markDead` is called jsut after
+      // a pool update, and in such case we will add to the dead
+      // list a node that no longer exist. The following check verify
+      // that the connection is still part of the pool before
+      // marking it as dead.
+      for (var i = 0; i < this.size; i++) {
+        if (this.connections[i].id === id) {
+          this.dead.push(id)
+          break
+        }
+      }
     }
     connection.status = Connection.statuses.DEAD
     connection.deadCount++
--- a/test/unit/connection-pool.test.js
+++ b/test/unit/connection-pool.test.js
@@ -74,6 +74,14 @@ test('API', t => {
     }, 10)
   })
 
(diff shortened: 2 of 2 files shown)


test/unit/connection-pool.test.js
```diff
@@ -74,6 +74,14 @@ test('API', t => {
...
   })
 
+  t.test('markDead should ignore connections that no longer exists', t => {
+    const pool = new ConnectionPool({ Connection, sniffEnabled: true })
+    pool.addConnection('http://localhost:9200/')
+    pool.markDead({ id: 'foo-bar' })
+    t.deepEqual(pool.dead, [])
+    t.end()
+  })
+
```
penge, line 84: Should be the same tested for `markAlive`?
Is it ok to `markAlive` a connection that isn't in the list of `connections`?

lib/pool/ConnectionPool.js
```diff
@@ -60,7 +60,17 @@ class ConnectionPool extends BaseConnectionPool {
...
+      // a pool update, and in such case we will add to the dead
+      // list a node that no longer exist. The following check verify
+      // that the connection is still part of the pool before
+      // marking it as dead.
+      for (var i = 0; i < this.size; i++) {
+        if (this.connections[i].id === id) {
+          this.dead.push(id)
+          break
+        }
+      }
```
penge, line 73: If not added to `dead`, it still continues on to set `.status`, `.deadCount`, `.resurrectTimeout`, or call `.sort()`.
Would it be better to not do these and return `this` right after?

test/unit/connection-pool.test.js
delvedor: There's no need to add it there, as both `markAlive` and `markDead` are just modifying the given connection instance.
The only place where `markDead` is actually doing something "dangerous", is when updating the `dead` array, which is just a cache of the dead nodes.

I'm checking the existence of a connection only when modifying the `dead` array to skip useless loops :)

lib/pool/ConnectionPool.js
delvedor: That's true, but it's not a problem as the connection will be garbage collected at the end of the function :)
There is no need to add an "exists" logic just to add an early return imho.

PR #1180, opened 2020-04-30
20200430 aws sniff adaptor

TheChrisOneil:


lib/Transport.js
```diff
@@ -386,7 +386,15 @@ class Transport {
 
       debug('Sniffing ended successfully', result.body)
       const protocol = result.meta.connection.url.protocol || 'http:'
-      const hosts = this.connectionPool.nodesToHost(result.body.nodes, protocol)
+      var url = ''
+      if (Object.prototype.hasOwnProperty.call(result.meta.connection, 'url') &&
+          typeof result.meta.connection.url === 'string') {
```
penge, line 391: This can be:

```js
if (typeof result.meta.connection.url === "string") {
  // ...
}
```

If it is not there, then it is `undefined`, therefore not a `string`.

lib/Transport.js
```diff
@@ -386,7 +386,15 @@ class Transport {
...
+        url = result.meta.connection.url
+      } else if ((Object.prototype.hasOwnProperty.call(result.meta.connection, 'url') &&
+          typeof result.meta.connection.url === 'object')) {
```
penge, line 394: This is enough:

```js
if (typeof result.meta.connection.url === "object") {
  // ...
}
```

PR #1231, opened 2020-06-24
Added option to run the benchmarks sequentially or concurrently

delvedor:
We can now pick which style to use with the `run` option, it can either be `sequential` or `concurrent`.


test/benchmarks/lib.js
```diff
@@ -119,8 +119,10 @@ class Action {
     assert(opts.action, 'Missing action name')
     assert(opts.action, 'Missing action category')
     assert(opts.measure, 'Missing action measure function')
+    assert(opts.run, 'Missing actionrun type (sequential or concurrent)')
```
karmi, line 122:
```suggestion
    assert(opts.run, 'Missing action run type (sequential or concurrent)')
```

PR #1239, opened 2020-06-30
Export a kibana restricted type definition

delvedor:
While working on #1243 and https://github.com/elastic/kibana/pull/69905 we have decided that we should start shipping a restricted type definition file for Kibana, which hides the snake_cased and callbacks style APIs.

The type can be used as follows:
```ts
import { KibanaClient } from '@elastic/elasticsearch/api/kibana'

const client: KibanaClient = new Client({
  node: 'http://localhost:9200'
})
```


scripts/utils/generateMain.js
```diff
@@ -177,11 +181,23 @@ function toPascalCase (str) {
   return str[0].toUpperCase() + str.slice(1)
 }
 
-function buildMethodDefinition (api, name, hasBody) {
+function buildMethodDefinition (opts, api, name, hasBody) {
   const Name = toPascalCase(name)
   const bodyType = ndjsonApiKey.includes(Name) ? 'RequestNDBody' : 'RequestBody'
   const defaultBodyType = ndjsonApiKey.includes(Name) ? 'Record<string, any>[]' : 'Record<string, any>'
 
+  if (opts.kibana) {
```
restrry, line 189: It looks like the script requires at least some minimal test-set. Do you need any help with this?

delvedor: The generated code is tested with the integration test and the types test.
The types test will now fail because the kibana output exposes less properties, which is expected.
I’m not entirely sure if it makes sense to write a test specifically for the script.

restrry: > The generated code is tested with the integration test and the types test. 
> The types test will now fail because the kibana output exposes less properties, which is expected.

This is good as long as Kibana-specific types are covered. Can we add a separate test for them?

scripts/generate.js
```diff
@@ -58,23 +59,33 @@ function start (opts) {
...
     )
 
-    const oldTypeDefString = readFileSync(typeDefFile, 'utf8')
-    const start = oldTypeDefString.indexOf('/* GENERATED */')
-    const end = oldTypeDefString.indexOf('/* /GENERATED */')
-    const newTypeDefString = oldTypeDefString.slice(0, start + 15) + '\n' + types + '\n  ' + oldTypeDefString.slice(end)
+    let oldTypeDefString = readFileSync(typeDefFile, 'utf8')
+    let start = oldTypeDefString.indexOf('/* GENERATED */')
+    let end = oldTypeDefString.indexOf('/* /GENERATED */')
+    let newTypeDefString = oldTypeDefString.slice(0, start + 15) + '\n' + types + '\n  ' + oldTypeDefString.slice(end)
```
restrry, line 72: nit: it's easy to forget to update length if we change **start** token. Do you think you could you refer it explicitly?
```js
const startToken = '/* GENERATED */';
let startPosition = oldTypeDefString.indexOf(startToken)
let newTypeDefString = oldTypeDefString.slice(0, start + startToken.length) + ...
```

delvedor: I don't expect it to change, also, we could still forget to update it here :D

sebelga: @delvedor Jumping in on @restrry comment. The recommended change is not only to cover the case where "start token" would change, but also to avoid "magic" number as per our styleguide (https://github.com/elastic/kibana/blob/master/STYLEGUIDE.md#magic-numbersstrings).
When I read `15` it does not mean anything, when I read `startToken.length` it is explicit what it is. Just my 2 cents 😊

PR #1243, opened 2020-07-02
Kibana client support

delvedor:
Proposal for offering first-class support to Kibana.

## Requirements
- [x] Avoid the creation of a wrapper
- [x] Some properties and methods of the client should not be accessed by mistake
  - [x] `transport`
  - [x] `connectionPool`
  - [x] `serializer`
  - [x] `close`
  - [x] `child`
  - [x] `extend`
  - [x] `on`
  - [x] `once`
  - [x] `emit`
- [ ] Simplified type definition *(the work done in https://github.com/elastic/elasticsearch-js/pull/1239 will merged here)*
- [ ] Ship the kibana type definition with the client
- [ ] The chosen solution will be part of the testing suite of the client
- [ ] Support of jest mocks

## Implementation
Currently, the idea is that the Kibana client will extend the base Client and hide the properties that should not be accessed by plugin users.
The main counterargument of this approach is that if Kibana core needs to access these methods, we'll need to re-expose them via symbols, for example:
```js
const kClose = Symbol('kibana-client-close')
class KibanaClient extends Client {
  get [kClose] () {
    return super.close
  }
}

const client = new KibanaClient()
client[kClose]()
```

A different solution is to exclude the properties we don't want to be accessed with a builder function.
```js
const client = new Client(opts)
function buildKibanaClient () {
  return {
    ...client,
    transport: {
      request: client.transport.request.bind(client.transport)
    },
    connectionPool: undefined,
    serializer: undefined,
    close () {
      throw new errors.ElasticsearchClientError('Cannot access close method')
    }
  }
}

const kclient = buildKibanaClient()
```

In my opinion, the first method is more elegant, as it uses idiomatic JavaScript to solve the issue, but will require you to use symbols for accessing "private" APIs. The second method feels hackier, as we'll need to bind the client in some places, but it removes the need to use Symbols at all.

@elastic/kibana-platform what do you think?

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,12 +21,15 @@ const { ConfigurationError } = errors
 const kInitialOptions = Symbol('elasticsearchjs-initial-options')
 const kChild = Symbol('elasticsearchjs-child')
 const kExtensions = Symbol('elasticsearchjs-extensions')
+const kTransport = Symbol('elasticsearchjs-transport')
+const kConnectionPool = Symbol('elasticsearchjs-connection-pool')
+const kSerializer = Symbol('elasticsearchjs-serializer')
+const kEventEmitter = Symbol('elasticsearchjs-event-emitter')
 
 const buildApi = require('./api')
 
-class Client extends EventEmitter {
+class Client {
   constructor (opts = {}) {
-    super()
     if (opts.cloud) {
       const { id, username, password } = opts.cloud
       // the cloud id is `cluster-name:base64encodedurl`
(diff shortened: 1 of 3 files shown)


test/unit/kibana-client.test.js
```diff
@@ -0,0 +1,312 @@
...
+    throw new errors.ElasticsearchClientError('Cannot access once method')
+  }
+}
+
+/**
+ * builder method
+ * pros:
+ *   - no need to use symbols in kibana core for accesssing properties
+ * cons:
+ *   - hackier solution, it requires a lot of small adjustments for binding the client instance
```
restrry, line 69: yeah. that looks rather fragile.

index.js
```diff
@@ -146,6 +154,30 @@ class Client extends EventEmitter {
     })
   }
 
+  get transport () {
+    return this[kTransport]
```
restrry, line 158: it can be as simple as `_transport` instead of using `Symbol`. not 100% safe, but Symbol doesn't provide full safety either.

delvedor:
Closing in favor of https://github.com/elastic/elasticsearch-js/pull/1239 for now as we decided to start simple and only ship a restricted type definition file.

PR #1254, opened 2020-07-09
 Allow the client name to be a symbol

delvedor:
As titled.

Related: https://github.com/elastic/kibana/pull/71228


lib/Transport.d.ts
```diff
@@ -52,7 +52,7 @@ export interface RequestEvent<TResponse = Record<string, any>, TContext = unknow
   warnings: string[] | null;
   meta: {
     context: TContext;
-    name: string;
+    name: string | Symbol;
```
restrry, line 55: Kibana bans `Symbol` usage https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md which could lead to problems with a facade around the ES client

PR #1256, opened 2020-07-13
Add support for a global context option

delvedor:
As titled.


index.d.ts
```diff
@@ -90,6 +90,7 @@ interface ClientOptions {
   generateRequestId?: generateRequestIdFn;
   name?: string;
   auth?: BasicAuth | ApiKeyAuth;
+  context?: Record<string, any>;
```
restrry, line 93: prefer to use `unknown`. it will enforce type casting before reading https://mariusschulz.com/blog/the-unknown-type-in-typescript

test/types/client-options.test-d.ts
```diff
@@ -616,3 +616,20 @@ expectError<errors.ConfigurationError>(
...
+  new Client({
+    node: 'http://localhost:9200',
+    context: { hello: 'world' }
+  })
+)
+
+expectError<errors.ConfigurationError>(
+  new Client({
+    node: 'http://localhost:9200',
+    context: 'hello world'
```
restrry, line 633: could you add a test for `context` reading within a `response` callback?

delvedor: It's already present in `test/types/client.test-d.ts` :)

PR #1263, opened 2020-07-20
Fix errors in type definitions

delvedor:
- The `ignore` paramater is part of the request options and not the request params.
- In the rest-api-spec boolean are written as string (`'true'` and `false'`), this pr generates them as `boolean`.

The type definition fix will be generated once in `master`, `7.x`, and `7.9`.
Related: https://github.com/elastic/kibana/pull/72388 https://github.com/elastic/kibana/pull/72289

### DIFF
--- a/scripts/utils/generateRequestTypes.js
+++ b/scripts/utils/generateRequestTypes.js
@@ -26,7 +26,6 @@ import { RequestBody, RequestNDBody } from '../lib/Transport'
 
 export interface Generic {
   method?: string;
-  ignore?: number | number[];
   filter_path?: string | string[];
   pretty?: boolean;
   human?: boolean;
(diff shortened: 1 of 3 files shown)


scripts/utils/generateRequestTypes.js
```diff
@@ -118,7 +117,31 @@ export interface ${toPascalCase(name)}${body ? `<T = ${bodyGeneric}>` : ''} exte
...
+              return `'${k}'`
+            }
+          })
+          .filter(k => {
+            if (foundTrue && foundFalse && (k === true || k === false)) {
+              return false
+            }
+            return true
+          })
+        if (foundTrue && foundFalse) {
```
restrry, line 141: I can imagine an endpoint with `false | 'some_string'` parameter

delvedor: I know, I'm being overcautious here :D

delvedor:
> are you going to release it as rc2?

Yes, I'll release as many as we need until we solve https://github.com/elastic/kibana/issues/70431.

> are we still waiting for the ES team response on the deleteByNamespace interface?

I've just asked, but we can merge this one in the meantime already.

PR #1293, opened 2020-08-31
Correct param name for `scroll` calls

IanSavchenko:
Not sure if this is an issue with previous version of the client but on 7.x there is definitely a mistake in the docs.

### DIFF
--- a/docs/examples/scroll.asciidoc
+++ b/docs/examples/scroll.asciidoc
@@ -102,7 +102,7 @@ async function run () {
     // get the next response if there are more quotes to fetch
     responseQueue.push(
       await client.scroll({
-        scrollId: body._scroll_id,
+        scroll_id: body._scroll_id,
         scroll: '30s'
       })
     )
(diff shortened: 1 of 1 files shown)


delvedor (commented):
The client does accept both snake_case and camelCase query parameters. Is the snippet not working for you?

IanSavchenko:
Ok, I didn't know that. Then I guess it works, I was surprised that all examples on the Internet actually have `scollId`. 

The reason I thought it was wrong was because TypeScript compiler had been complaining about it, insisting that it should be `scroll_id`.

delvedor:
@IanSavchenko Yes, the types definitions only accept snake_case, as it's not possible to enforce a required parameter if we support both styles.

IanSavchenko:
@delvedor Thank you for the answer! I can think of ways making both cases working with TypeScript but it will be messy and hardly worth it.

Ok, then I leave it up to you if want to merge this.

delvedor:
I'm planning to migrate the whole docs to snake_case, but I would like to do it in a single pr :)
Thank you for contributing!

PR #1299, opened 2020-09-04
Adding trailing slash in request path to avoid ECONNRESET on some spe…

ncastay:
…cific network conditions / machine setup

While using legacy elasticsearch js client in a project, I faced some issue leading to having ECONNRESET when querying ES.
Not all queries, but some of them (for example _count was OK while _search ended up closing the connection right away).

Found out it was linked to having or not a trailing slash in the requested path URL.

And there is already an issue in the legacy client about that :
https://github.com/elastic/elasticsearch-js-legacy/issues/1

Although it states that the new client wont have this problem because the source code is handling the path creation in a different way, I found out after upgrading that it was changed again since that time and the problem is still there on latest source code.

Hence this pull request adding '/' on all paths used.
I'm not sure which paths can be linked to the issue or if it would be the same ones for different network configs / machine so I added to all of them to be safe.
I don't think there should be any harm in having it (even if I would agree that not having it shouldn't cause issue in the first place...)

Let me know if you believe another implementation for fixing this would be better.

Best regards,
  Nicolas

### DIFF
--- a/api/api/async_search.delete.js
+++ b/api/api/async_search.delete.js
@@ -75,7 +75,7 @@ function buildAsyncSearchDelete (opts) {
     var path = ''
 
     if (method == null) method = 'DELETE'
-    path = '/' + '_async_search' + '/' + encodeURIComponent(id)
+    path = '/' + '_async_search' + '/' + encodeURIComponent(id) + '/'
 
     // build request object
     const request = {
--- a/api/api/async_search.get.js
+++ b/api/api/async_search.get.js
@@ -79,7 +79,7 @@ function buildAsyncSearchGet (opts) {
     var path = ''
 
     if (method == null) method = 'GET'
-    path = '/' + '_async_search' + '/' + encodeURIComponent(id)
+    path = '/' + '_async_search' + '/' + encodeURIComponent(id) + '/'
 
     // build request object
     const request = {
--- a/api/api/async_search.submit.js
+++ b/api/api/async_search.submit.js
@@ -140,10 +140,10 @@ function buildAsyncSearchSubmit (opts) {
 
     if ((index) != null) {
       if (method == null) method = 'POST'
-      path = '/' + encodeURIComponent(index) + '/' + '_async_search'
+      path = '/' + encodeURIComponent(index) + '/' + '_async_search' + '/'
     } else {
       if (method == null) method = 'POST'
-      path = '/' + '_async_search'
+      path = '/' + '_async_search' + '/'
     }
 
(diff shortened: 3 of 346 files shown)


ncastay:
I signed the CLA now and it seems I forgot to upgrade the unit tests accordingly, I'll add that.

delvedor (changes requested):
Hello! I'm not sure I follow, can you provide ane example to reproduce the issue you are having?

ncastay:
@delvedor

Unfortunately no...
I have found out that depending on your environment you can have this issue, however I cannot yet pinpoint its root cause.

We have 2 machines with the exact same node application deployed, and our ES instance is on a third one (with both machines accessing it).
From machine 1 everything works fine, but from machine 2 I cannot access some ES endpoints unless I input a trailing slash on the request.

`
[root@machine2 ~]$ curl -X GET http://es.instance.com:9200/index-tag/project/_search
curl: (56) Recv failure: Connection reset by peer
[root@machine2 ~]$ curl -X GET http://es.instance.com:9200/index-tag/project/_search/
{"took":2,"timed_out":false,"_shards":{"total":3,"successful":3,"skipped":0,"failed":0},"hits":{"total":7214,"max_score":1.0,"hits":,...
`

The problem happens on some endpoints but not others (i can access _count or _mapping for example, but not _search).

So far I could not figure out what's the root cause of this, there must be some environment specificity on this machine causing the problem, however I figured the change would only make the elasticsearch package more robust to various environments, so it shouldn't hurt.

Best regards,
  Nicolas

delvedor (commented):
I've checked with my peers inside the company and we never had such an issue, so very likely is dependent on your infrastructure, do you have a proxy between the client and Elasticsearch?

I'll not merge this pr, as Elasticsearch does not need trailing slashes.
What you can do for fixing your issue, is to create a custom Transport class that does this for you, eg:
```js
const { Client, Transport } = require('@elastic/elasticsearch')

class MyTransport extends Transport {
  request (params, options, callback) {
    if (!params.path.endsWith('/')) {
       params.path += '/'
    }
    return super.request(params, options, callback)
  }
}

const client = new Client({
  Transport: MyTransport
})
```

Thank you for contributing!

ncastay:
OK, I've tried the custom Transport and it's working alright now, so I'll just go with that.

Yesterday I thought it could be linked to having a very old libcurl version (as it was one of the difference I could find between the 2 machines), but upgrading it didn't help with the issue.

Thank you for the "workaround".

PR #1300, opened 2020-09-04
Fix RequestOptions.body type to include null

### LINKED ISSUE #1291: RequestOptions type says body can't be null but it can
## 🐛 Bug Report

This is the definition on `Connection.d.ts`:
```typescript
interface RequestOptions extends http.ClientRequestArgs {
  asStream?: boolean;
  body?: string | Buffer | ReadableStream;
  querystring?: string;
}
```

## To Reproduce

I've been using this class to provide authenticated connection with AWS Elasticsearch:

```typescript
import { Readable } from 'stream';
import { Connection as UnsignedConnection } from '@elastic/elasticsearch';
import * as AWS from 'aws-sdk';
import RequestSigner from 'aws-sdk/lib/signers/v4';
import { ClientRequest, IncomingMessage } from 'http';

class AwsElasticsearchError extends Error {}
type RequestOptions = Parameters<UnsignedConnection['request']>[0];

class AwsSignedConnection extends UnsignedConnection {
  public request(
    params: RequestOptions,
    callback: (err: Error | null, response: IncomingMessage | null) => void,
  ): ClientRequest {
    const signedParams = this.signParams(params);
    return super.request(signedParams, callback);
  }

  private signParams(params: RequestOptions): RequestOptions {
    const region = AWS.config.region || process.env.AWS_DEFAULT_REGION;
    if (!region) throw new AwsElasticsearchError('missing region configuration');
    if (!params.method) throw new AwsElasticsearchError('missing request method');
    if (!params.path) throw new AwsElasticsearchError('missing request path');
    if (!params.headers) throw new AwsElasticsearchError('missing request headers');

    const endpoint = new AWS.Endpoint(this.url.href);
    const request = new AWS.HttpRequest(endpoint, region);

    request.method = params.method;
    request.path = params.querystring
      ? `${params.path}/?${params.querystring}`
      : params.path;
    if (typeof params.body === 'string') request.body = params.body;
    else if (params.body instanceof Buffer) request.body = params.body.toString();
    else if (params.body instanceof Readable) throw new AwsElasticsearchError('unsupported');
    else if (params.body !== undefined) {
      throw new AwsElasticsearchError('body type unhandled');
    }

    Object.entries(params.headers).forEach(([header, value]) => {
      if (typeof value === 'string') request.headers[header] = value;
      else if (typeof value === 'number') request.headers[header] = `${value}`;
      else if (Array.isArray(value)) request.headers[header] = value.join('; ');
      else if (value !== undefined) throw new AwsElasticsearchError('header type unhandled');
    });
    request.headers.Host = endpoint.host;

    const signer = new RequestSigner(request, 'es');
    signer.addAuthorization(AWS.config.credentials, new Date());
    return request;
  }
}
```

When using the connection, it throws with `body type unhandled` because `params.body` is `null`.

## Expected behavior

The typing should say it can be null, or it should never be null.

## Your Environment

- *node version*: 10.14.1
- `@elastic/elasticsearch` *version*: >=7.9.0
- *os*: Linux

villasv:
Fixes #1291

### DIFF
--- a/lib/Connection.d.ts
+++ b/lib/Connection.d.ts
@@ -51,7 +51,7 @@ interface ConnectionRoles {
 
 interface RequestOptions extends http.ClientRequestArgs {
   asStream?: boolean;
-  body?: string | Buffer | ReadableStream;
+  body?: string | Buffer | ReadableStream | null;
   querystring?: string;
 }
 


delvedor (approved):
LGTM

PR #1311, opened 2020-09-16
Fix rendering of name field in configuration.md

kellycampbell:
Name field had `string | symbol` but the | was treated as a table column separator.


kellycampbell:
@delvedor

delvedor (approved):
Thanks!

PR #1312, opened 2020-09-17
Fixes 'Documentation' link from indices.getAlias

rudolf:
The 'Documentation' link for `indices.getAlias` was pointing to the update aliases documentation not get aliases documentation.


delvedor (commented):
Thanks for the pr! I can't merge it, because this file is autogenerated from the [JSON spec](https://github.com/elastic/elasticsearch/tree/master/rest-api-spec/src/main/resources/rest-api-spec/api). You should update the URL over there.

Issue #1321, opened 2020-09-23
searchScroll helper don't clear the scroll cursor after looping over all documents


abarre:
## 🐛 Bug Report

After the [searchScroll](https://github.com/elastic/elasticsearch-js/blob/7b11152a128525644208b317ed073c2a38353062/lib/Helpers.js#L73) has loop over all documents, the scroll cursor is not released.

The [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/search-request-scroll.html#_clear_scroll_api) states : 

> Search context are automatically removed when the scroll timeout has been exceeded. However keeping scrolls open has a cost, as discussed in the previous section so scrolls should be explicitly cleared as soon as the scroll is not being used anymore using the clear-scroll API

Current behavior leak the search context for the scroll window (1 minute by default).

## Expected behavior

After this [while loop](https://github.com/elastic/elasticsearch-js/blob/7b11152a128525644208b317ed073c2a38353062/lib/Helpers.js#L106), I think that there is a missing await clear(): 

```diff 
while (response.body.hits && response.body.hits.hits.length > 0) {
...
}
+ await clear();
```

I will be happy to make a PR if the expected behavior is confirmed.

## Your Environment

- `@elastic/elasticsearch` *version*: 7.9.1

delvedor:
Hello! You are right, great catch!

> I will be happy to make a PR if the expected behavior is confirmed.

Please do! Remember to update the test as well, thanks!

abarre:
Hi @delvedor, 

when do you plan to release a patch version with this fix for the branch 7.x ?

Thanks

PR #1322, opened 2020-09-24
[#1321] Fix scroll cursor leak inside searchScroll helper.

abarre:
I didn't find a way to test this change.
The leak is not visible in the client side but only in the server side during the scroll window.

### DIFF
--- a/lib/Helpers.js
+++ b/lib/Helpers.js
@@ -127,6 +127,10 @@ class Helpers {
         throw new ResponseError(response)
       }
     }
+
+    if (!stop && scroll_id) {
+      await clear()
+    }
   }
 
   /**


delvedor (changes requested):
Thanks for the fix!
Regarding the test, you should add a new test case in [this file](https://github.com/elastic/elasticsearch-js/blob/master/test/unit/helpers/scroll.test.js) to verify that the last call (if the conditions are met) is a clear scroll.

abarre:
I will modify this PR in the next days. I had a quick look at the code and I see that the [test code](https://github.com/elastic/elasticsearch-js/blob/b8940533a915acedbfac0f11ebf645dad795a2a0/test/unit/helpers/scroll.test.js#L33) expect that the last request doesn't send a scroll_id in the response.
Can you confirm that this is a mistake and that in real life, all responses contain a scroll id ?

I didn't find the documentation on this point.

delvedor:
I've checked and the `scroll_id` is always present in the body, but it can change over time.
I've opened https://github.com/elastic/elasticsearch-js/pull/1331 with the fix and the updated test.
Closing this, thank you a lot for contributing!

PR #1332, opened 2020-10-07
The agent function should take the Connection contructor options as argument

### LINKED ISSUE #1330: Client#constructor's agent function doesn't designate the protocol
Currently, the `Client#constructor` allows the `agent` option to specify a function that returns an agent. However, when the `agent` function is called, it has no arguments, so it's not possible to determine whether or not a `http.Agent` or ` https.Agent` should be returned.

I'd like to use this in Kibana, to ensure that all instances of the Client share an Agent, when possible, to efficiently reuse sockets. However, Kibana allows users to specify both HTTP and HTTPs nodes, so an Agent itself can't be passed into the `Client#constructor`. 

The legacy elasticsearch-js client is providing an instance of the `HttpConnector` as the first argument to the `createNodeAgent` function, which makes this possible.

delvedor:
As titled.

Closes #1330

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -149,7 +149,9 @@ const client = new Client({
 
 const client = new Client({
   node: 'http://localhost:9200',
-  agent: () => new CustomAgent()
+  // the function takes as parameter the option
+  // object passed to the Connection constructor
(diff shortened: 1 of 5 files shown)


kobelb (approved):
This is perfect, thanks so much!

Issue #1088, opened 2020-02-12
DeserializationError: Unexpected end of JSON input


msisti:
## 🐛 Bug Report

We are making Elasticsearch queries in a test file. The test file has multiple tests, each performing a query. We tried reordering the tests in the file and found that the first test in the file will be the only one that fails. Even with a stripped down body for the query, we are still inconsistently getting the following error:

![image](https://user-images.githubusercontent.com/10967659/74376299-d8321d80-4daf-11ea-8e09-ce1e973c7875.png)
Text: "DeserializationError: Unexpected end of JSON input"

## To Reproduce

Run the following test multiple times (it takes a variable number of attempts to get the failure):

```js
import { Client } from '@elastic/elasticsearch

const defaultClient = new Client({
    node: host,
    sniffOnStart: true
})

describe('elastic test', () => {
    test('it should not have deserialization errors', async () => {
        await defaultClient.search({
            index: 'common',
            type: '_doc',
            body: {
                query: {
                    match_all: {}
                }
            }
        })
    })
})
```

## Expected behavior

The test should be passing consistently.

## Your Environment

- *node version*: 10.11.0
- `@elastic/elasticsearch` *version*: ^7.5.1
- *os*: Mac
- *Elasticsearch instance version*: 7.4.2

Please let me know if you need any additional information!

delvedor:
Hello!
Unfortunately, the test you have provided does not help to understand what the issue is.
Given that the error is a serialization error, it means that the response body is an invalid JSON.
For figure out what's going wrong, you have two ways:
1. Enable debug mode:
```sh
DEBUG=elasticsearch node test.js
```

2. Use a custom serializer and inspect the JSON string:
```js
const { Client, Serializer, errors } = require('@elastic/elasticsearch')

class DebugSerializer extends Serializer {
  deserialize (json) {
    try {
      var object = JSON.parse(json)
    } catch (err) {
      console.dir(json) // add a log here
      throw new errors.DeserializationError(err.message)
    }
    return object
  }
}

const client = new Client({
  node: 'http://localhost:9200',
  Serializer: DebugSerializer
})

client.search({
  index: 'test',
  body: {
    query: {
      match_all: {}
    }
  }
}, (err, res) => {
  console.log(err || res)
})
```

Both methods are fine, the first will print many logs in the console, so you might want to use the second.

msisti:
[...]

Issue #1297, opened 2020-09-02
AbortController/AbortSignal support


lukasolson:
Currently, the `TransportRequestPromise` type is an extension of the built-in `Promise` type, which contains an `abort` method to programmatically cancel the request. Extending the native `Promise` makes it difficult to chain/wrap or use `async`/await`.

As an alternative (or additional) approach, I'd like to suggest using the `AbortController`/`AbortSignal` approach used natively (supported in most modern browsers, polyfill available) by `fetch`:

```ts
const controller = new AbortController();

const response = await esClient.search(params, {
  signal: controller.signal
});

// Then, if we need to abort for some reason, maybe in an event handler:
controller.abort();
```

It sounds like @delvedor has been considering this since it is likely to land natively in Node [1]. From our conversations:

> That’s something I’ve been thinking for a while, as a similar api is landing to node core.
Unfortunately we can’t drop the support for the .abort method as it would be a breaking change, but we can support both and deprecate the old one. I’ve already implemented it in userland libraries (https://github.com/nodejs/undici), the main reason why the client has an .abort method, is because the legacy one had it. I want to see how Node.js will end up supporting this, to avoid creating an api not compatible with the rest of the ecosystem.

[1] https://github.com/nodejs/node/pull/33527

PR #1343, opened 2020-11-03
Handle connectivity issues while reading the body

delvedor:
Until now the client was not verifying the `content-length` header, this pr address this issue.
For supporting this, the unzip process has been moved from the `Connection` class to the `Transport` class.

A byproduct of this pr is the fix of #1088.

Closes: #1088 
/cc: @msisti @tomsquest

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -25,7 +25,6 @@ const hpagent = require('hpagent')
 const http = require('http')
 const https = require('https')
 const debug = require('debug')('elasticsearch')
-const decompressResponse = require('decompress-response')
 const pump = require('pump')
 const INVALID_PATH_REGEX = /[^\u0021-\u00ff]/
 const {
@@ -83,7 +82,6 @@ class Connection {
 
   request (params, callback) {
     this._openRequests++
-    var ended = false
 
     const requestParams = this.buildRequestObject(params)
     // https://github.com/nodejs/node/commit/b961d9fd83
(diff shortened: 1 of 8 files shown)


delvedor:
It looks like there is a problem with Node v14, looking into it.

lib/Connection.js
```diff
@@ -96,53 +94,38 @@ class Connection {
...
-        ended = true
-        this._openRequests--
-        callback(new ConnectionError(err.message), null)
-      }
-    })
+    const onTimeout = () => {
+      cleanListeners()
+      this._openRequests--
+      request.once('error', () => {}) // we need to catch the request aborted error
+      request.abort()
```
rudolf, line 107: If you use `request.destroy()` it might solve the v14 problem and it will also not emit an error event so you will no longer need the empty error handler.

rudolf (approved):
Left a small comment, but otherwise looks good to me

lib/Connection.js
delvedor: Thanks for the review! I'm looking into this, as `request.abort` has been deprecated in newer versions of node.
Very likely I'll address this in a separate pr while working on https://github.com/elastic/elasticsearch-js/issues/1297.

delvedor:
Thanks to the awesome [Node.js contributors](https://github.com/nodejs/node/issues/35958) I've discovered the `response.on('aborted')` event, which among other things, addresses specifically this issue. I've removed the content length check as the aborted event always supersede it.
The best part is that it works in the same way from Node.js 10 and above.

lib/Transport.js
```diff
@@ -237,9 +237,9 @@ class Transport {
         if (isCompressed) {
           payload = Buffer.concat(payload)
         }
-        if (result.headers['content-length'] !== undefined && Number(result.headers['content-length']) !== payload.length) {
+        if (result.headers['content-length'] !== undefined && Number(result.headers['content-length']) !== Buffer.byteLength(payload)) {
```
Spacefish: That´s why everyone loves javascript 👯‍♀️

Spacefish:
We have the same issue.. I noticed something strange... For all requests failing the _source: reply from ES to Kibana contains a 3 bytes UTF-8 Bom inside the JSON string.. not sure if this is even legal:
image

SSANSH:
hi, possible to back port this fix on 5.x ?

PR #1346, opened 2020-11-06
Added node.js support doc

delvedor:
Since the client **must** follow the stack versioning, it's really hard to follow the [node.js release cadence](https://nodejs.org/en/about/releases/), so we have decided to allow dropping EOL node.js version in minor releases. This pr contains the formalization of the process we will follow in the future.

- [x] asciidoc updated
- [x] README updated

### DIFF
--- a/docs/installation.asciidoc
+++ b/docs/installation.asciidoc
@@ -37,8 +37,8 @@ to support that version for at least another minor release. If you are using the
 with a version of Node.js that will be unsupported soon, you will see a warning
 in your logs (the client will start logging the warning with two minors in advance).
 
-Because of the reasons defined above, unless you are *always* using a supported
-version of Node.js, we recommend defining the client dependency in your
+Unless you are *always* using a supported version of Node.js, 
+we recommend defining the client dependency in your
 `package.json` with the `~` instead of `^`. In this way, you will lock the
 dependency on the minor release and not the major. (for example, `~7.10.0` instead
 of `^7.10.0`).
--- a/docs/installation.asciidoc
+++ b/docs/installation.asciidoc
@@ -40,7 +40,7 @@ in your logs (the client will start logging the warning with two minors in advan
 Because of the reasons defined above, unless you are *always* using a supported
 version of Node.js, we recommend defining the client dependency in your
 `package.json` with the `~` instead of `^`. In this way, you will lock the
-dependency on the minor release and not the major. (eg: `~7.10.0` instead
+dependency on the minor release and not the major. (for example, `~7.10.0` instead
 of `^7.10.0`).
 
 [%header,cols=3*]
--- a/docs/installation.asciidoc
+++ b/docs/installation.asciidoc
@@ -26,7 +26,7 @@ To learn more about the supported major versions, please refer to the
 
 NOTE: The minimum supported version of Node.js is `v8`.
 
-The client versioning follows the Elastic Stack versioning, this means that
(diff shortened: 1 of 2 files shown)


philkra (approved):
LGTM

docs/installation.asciidoc
```diff
@@ -20,13 +20,48 @@ npm install @elastic/elasticsearch@<major>
 To learn more about the supported major versions, please refer to the 
 <<js-compatibility-matrix>>.
 
+[discrete]
+[[nodejs-support]]
+=== Node.js support
+
+NOTE: The minimum supported version of Node.js is `v8`.
+
+The client versioning follows the Elastic Stack versioning, this means that
```
szabosteve, line 29:
```suggestion
The client versioning follows the {stack} versioning, this means that
```

docs/installation.asciidoc
```diff
@@ -20,13 +20,48 @@ npm install @elastic/elasticsearch@<major>
...
+major, minor, and patch releases are done following a precise schedule that
+often does not coincide with the https://nodejs.org/en/about/releases/[Node.js release] times.
+
+For avoiding supporting insecure and unsupported versions of Node.js, the
```
szabosteve, line 33:
```suggestion
To avoid support issues and unsupported versions of Node.js, the
```

docs/installation.asciidoc
```diff
@@ -20,13 +20,48 @@ npm install @elastic/elasticsearch@<major>
...
+client *will drop the support of EOL versions of Node.js between minor releases*.
+Typically, as soon as a Node.js version goes into EOL, the client will continue
+to support that version for at least another minor release. If you are using the client
+with a version of Node.js that will be unsupported soon, you will see a warning
+in your logs (the client will start logging the warning with two minors in advance).
+
+Because of the reasons defined above, unless you are *always* using a supported
+version of Node.js, we recommend defining the client dependency in your
+`package.json` with the `~` instead of `^`. In this way, you will lock the
+dependency on the minor release and not the major. (eg: `~7.10.0` instead
```
szabosteve, line 43:
```suggestion
dependency on the minor release and not the major. (for example, `~7.10.0` instead
```

docs/installation.asciidoc
```diff
@@ -20,13 +20,48 @@ npm install @elastic/elasticsearch@<major>
...
+
+For avoiding supporting insecure and unsupported versions of Node.js, the
+client *will drop the support of EOL versions of Node.js between minor releases*.
+Typically, as soon as a Node.js version goes into EOL, the client will continue
+to support that version for at least another minor release. If you are using the client
+with a version of Node.js that will be unsupported soon, you will see a warning
+in your logs (the client will start logging the warning with two minors in advance).
+
+Because of the reasons defined above, unless you are *always* using a supported
+version of Node.js, we recommend defining the client dependency in your
```
szabosteve, line 41:
```suggestion
Unless you are *always* using a supported version of Node.js, 
we recommend defining the client dependency in your
```

szabosteve (approved):
Thank you for adding this section @delvedor. LGTM.
I left some very minor suggestions, mostly nits. Please feel free to take or leave them.

docs/installation.asciidoc
delvedor: I want to keep the "insecure", as using an EOL version is not safe :)

Mpdreamz (approved):
LGTM!

PR #1364, opened 2020-11-26
Added API name to request metadata

delvedor:
This is very useful if you need to filter logging and you don't want to use regular expressions.

Usage: 
```js
const client = new Client({
  node: 'http://localhost:9200'
})

client.on('response', (err, { meta }) => {
  if (err) {
    logger.error(err)
  } else {
    logger.info(`Api called: ${meta.request.options.api}`) // 'elasticsearch.info'
  }
})

client.info(console.log)
```

### DIFF
--- a/api/api/async_search.js
+++ b/api/api/async_search.js
@@ -55,6 +55,7 @@ AsyncSearchApi.prototype.delete = function asyncSearchDeleteApi (params, options
     querystring
   }
 
+  options.api = 'elasticsearch.async_search.delete'
   return this.transport.request(request, options, callback)
 }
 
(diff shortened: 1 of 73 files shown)


trentm:
I *like* this feature. I think it would make matching of ES API endpoints faster in the Node.js APM Agent. However, after some discussion with the APM team there was a desire to have the config for selecting ES API endpoints for request body capture be **shared between languages**. This "api" name isn't a common language between all the ES clients. The only common mechanism for matching ES API endpoints is the ES URL path, i.e. what the current Node.js APM Agent [is already using to match](https://github.com/elastic/apm-agent-nodejs/blob/5303ce58d358f138de1f0c8097029e9e7d220fcb/lib/instrumentation/modules/elasticsearch.js#L6).

So if the Node.js APM Agent was the only use case for this "api" name, you might want to consider not merging this for now.

PR #1365, opened 2020-11-30
Added new observability events

delvedor:
This pr adds two new events, `serialization` and `deserialization`.
The event order is described in the following graph, in some edge cases, the order is not guaranteed.
You can find in `test/acceptance/events-order.test.js` how the order changes based on the situation.

```
serialization
  │
  │ (serialization and compression happens between those two events)
  │
  └─▶ request
        │
        │ (actual time spent over the wire)
        │
        └─▶ deserialization
              │
              │ (deserialization and decompression happens between those two events)
              │
              └─▶ response
```

cc @trentm


EamonnTP:
@elasticmachine, run elasticsearch-ci/docs

PR #1367, opened 2020-12-02
Fix changelog display

delvedor:
As titled.


szabosteve (approved):
Thank you for the fix @delvedor! LGTM!

PR #1373, opened 2020-12-14
Added x-elastic-client-meta header

delvedor:
This pr add a new option named `enableMetaHeader `. If true, adds a header named `'x-elastic-client-meta'`, containing some minimal telemetry data, such as the client and platform version.

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,6 +33,8 @@ const Serializer = require('./lib/Serializer')
 const errors = require('./lib/errors')
 const { ConfigurationError } = errors
 const { prepareHeaders } = Connection.internals
+const clientVersion = require('./package.json').version
+const nodeVersion = process.versions.node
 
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -125,13 +127,18 @@ class Client extends ESAPI {
...
+        proxy: null,
+        enableMetaHeader: true
       }, opts)
 
     this[kInitialOptions] = options
     this[kExtensions] = []
     this.name = options.name
 
+    if (options.enableMetaHeader) {
+      options.headers['x-elastic-client-meta'] = `es=${clientVersion},js=${nodeVersion},hc=${nodeVersion},t=${clientVersion}`
```
sethmlarson, line 139: First three keys should be `es=, js=, t=` after that order doesn't matter

PR #1394, opened 2021-01-30
Add InferSearchResponse type

dgieselaar:
quick POC to explore how to add the inferred search response types to the elasticsearch-js repo


delvedor:
We have decided that this utility should live in Kibana until is stable enough, then we can move it directly in the client.
Closing for now!

PR #1400, opened 2021-02-08
[helpers] add support for transport options to all helpers

spalger:
I'm looking to specify custom headers while using the bulk helpers, which isn't currently currently supported in this helper. This change adds support for this param, extends it to include the `,h=bp` piece of the `x-elastic-client-meta` just like the `client.helpers.scroll()`, and passes the `reqOptions` to `client.bulk()`.

Also added support for `reqOptions` to `client.helpers.msearch()`. It's already supported in all other helpers


delvedor (commented):
This is very useful, thanks! We should add the same for other helpers.
If you don't have time to do it, I can do it soon.

delvedor (approved):
LGTM

PR #1402, opened 2021-02-18
Remove Node.js v8 support

delvedor:
As titled, according to our [support matrix](https://github.com/elastic/elasticsearch-js#nodejs-support).


delvedor:
Momentarily disabling `/api` folder linting, as the code generation will happen directly in the target branch.

PR #1412, opened 2021-03-08
Show how to disable secure json parsing

### LINKED ISSUE #1408: Query fails with "DeserializationError: Object contains forbidden prototype property" when the response object has a "constructor" key
## 🐛 Bug Report

Query fails with `DeserializationError: Object contains forbidden prototype property`.

This started happening when we upgraded from v7.3 to v7.11. This PR seems to be the cause: https://github.com/elastic/elasticsearch-js/pull/1110

## To Reproduce

Issue a query that results in `constructor` being an object key in the JSON response. In our case it was detected in a request to the [Term Vectors API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-termvectors.html), but I guess it can happen in many other places, like in an aggregation named `constructor`, etc. Example from the Term Vectors API response that affected us:

```
{
  "constructor": {
    "doc_freq": 1,
    "ttf": 1,
    "term_freq": 1
  }
}
```

See https://github.com/elastic/elasticsearch-js/pull/1110#issuecomment-790479130 for a full example.

## Expected behavior

No errors thrown from a perfectly valid response.

## Your Environment

- *node version*: 14
- `@elastic/elasticsearch` *version*: 7.11
- *os*: Linux

delvedor:
As titled.

Closes #1408

### DIFF
--- a/docs/advanced-config.asciidoc
+++ b/docs/advanced-config.asciidoc
@@ -118,3 +118,39 @@ const client = new Client({
   Serializer: MySerializer
 })
 ----
+
+[discrete]
+[[deserialization]]
+==== Deserialization
+
+By default, the deserialization is done with https://github.com/fastify/secure-json-parse[`secure-json-parse`]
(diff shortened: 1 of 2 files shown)


delvedor:
Superseded by https://github.com/elastic/elasticsearch-js/pull/1414.

Issue #1404, opened 2021-02-24
Usage link in README is dead


AmazingTurtle:
## 🐛 Bug Report

The link in the README.md is pointing to https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/client-usage.html which is dead (404).

delvedor:
Closed by #1413.

PR #1413, opened 2021-03-10
Update usage link (#1404)

khalilmansouri:
Update usage link in readme, current one seems to be dead (#1404)


delvedor (approved):
Thanks!

Issue #1393, opened 2021-01-25
.finally() not available on TransportRequestPromise


dgieselaar:
## 🐛 Bug Report

`.finally()` is not available on the promise returned by `.search()` (and possibly other calls).

## To Reproduce

Paste your code here:

```js
client.search(params).finally(doSomething)
```

Errors out:
`[12:57:08.510] [error][http] TypeError: promise.finally is not a function`

## Expected behavior

`.finally()` should be available on TransportRequestPromise, or the type should reflect that it's not.

## Your Environment

- *node version*: 14
- `@elastic/elasticsearch` *version*: 8.x canary
- *os*: Mac

PR #1415, opened 2021-03-10
fix(TransportRequestPromise): add finally method to TransportRequestP…

yahiakr:
Closes: #1393

### DIFF
--- a/lib/Transport.d.ts
+++ b/lib/Transport.d.ts
@@ -120,7 +120,7 @@ export interface TransportRequestCallback {
 
 export interface TransportRequestPromise<T> extends Promise<T> {
   abort: () => void;
-  finally(onfinally?: (() => void) | undefined | null): Promise<T>;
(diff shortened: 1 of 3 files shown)


delvedor (changes requested):
Hello! Thank you for sending a pr :)
Unfortunately, this change won't fix the issue, as you are changing only the type definition and not the original JavaScript code.

I think the best way to address this issue is to extend from `PromiseLike<T>` and add the catch definition inside the `TransportRequestPromise` interface.

yahiakr:
Hello @delvedor!

Thanks for your response & your guidance.
I fixed it. I’m adding the tests.

I will let you know once I complete it.

delvedor:
jenkins run the tests please

delvedor (approved):
LGTM

delvedor:
jenkins run the tests please

Issue #955, opened 2019-09-03
Typescript: Definition of _source parameter doest not allow boolean false


pierremalletneo9:
## 🐛 Bug Report

The typescript definition of the `_source` params is `_source?: string | string[];` 

```
export interface Search<T = any> extends Generic {
  index?: string | string[];
  type?: string | string[];
  ...
  _source?: string | string[];
  _source_excludes?: string | string[];
  _source_includes?: string | string[];
  ...
}
```

But it should allow boolean false value since the behavior of an Es request is totally different with `source: false` and `source: ''`. The first one strips the source deserialization by elasticsearch where the second need to deserialize the source of the documents and return nothing in the response. 

## Expected behavior

It should be possible to set `_source` to the boolean value false

## Your Environment

- *node version*: 10
- `@elastic/elasticsearch` *version*: >=7.3.0
- *os*: Linux

twixr:
I ran into the same issue. Trying to update a document with _source: false
```
Overload 1 of 4, '(params?: Update<{ doc: Record<string, unknown>; }> | undefined, options?: TransportRequestOptions | undefined): TransportRequestPromise<ApiResponse<Record<string, any>, unknown>>', gave the following error.
    Type 'false' is not assignable to type 'string | string[] | undefined'.
```

According to the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#docs-update-api-query-params):
```
_source
(Optional, list) Set to false to disable source retrieval (default: true). You can also specify a comma-separated list of the fields you want to retrieve.
```

delvedor:
This now has been fixed.

PR #1416, opened 2021-03-10
Allow _source to accept boolean in requestParams

khalilmansouri:
Based on [rest-spec-api](https://github.com/elastic/elasticsearch/tree/master/rest-api-spec/src/main/resources/rest-api-spec) `_source` should accept `boolean` beside `string` and `string[]`.
closes: #955


delvedor (commented):
Hello! Thank you for this pr :)
The file you are changing is autogenerated, and your fix will be overwritten at the next code generation.
This fix should be done in [`scripts/utils/generateRequestTypes.js`](https://github.com/elastic/elasticsearch-js/blob/master/scripts/utils/generateRequestTypes.js).

Given the work that is happening in https://github.com/elastic/elasticsearch-js/pull/1358, there is no need to send this fix, as the types will be generated in a completely different way in the future. Thank you for contributing!

PR #1457, opened 2021-04-26
Improve response error message

### LINKED ISSUE #1452: ResponseError provides root_cause of an exception
## 🚀 Feature Proposal

To refactor `ResponseError.toString` method format to include `root_cause` details provided by Elasticsearch server.

## Motivation

Applications often log a caught exceptions without additional effort for serialization.
If you log `ResponseError`
```js
try {
  await client.search(...);
} catch (e) {
  console.error(e)
}
```
you don't get an original `error_cause`:
```
    ResponseError: illegal_argument_exception
        at onBody (.../node_modules/@elastic/elasticsearch/lib/Transport.js:337:23)
        at IncomingMessage.onEnd (.../node_modules/@elastic/elasticsearch/lib/Transport.js:264:11)
        at IncomingMessage.emit (events.js:327:22)
        at endReadableNT (internal/streams/readable.js:1327:12)
        at processTicksAndRejections (internal/process/task_queues.js:80:21) {
      meta: {
        body: { error: [Object], status: 400 },
        statusCode: 400,
        headers: {
          'content-type': 'application/json;charset=utf-8',
          'content-length': '367'
        },
        meta: {
          context: null,
          request: [Object],
          name: 'elasticsearch-js',
          connection: [Object],
          attempts: 0,
          aborted: false
        }
      }
    }
```
that makes debugging quite painful.

To make`error.message` actionalbe,  it should include the `root_cause` provided by Elasticsearch server: `ResponseError: [illegal_argument_exception]. caused by [{ type: "...", reason: "..." }]`

delvedor:
Closes #1452.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -92,7 +92,7 @@ class ResponseError extends ElasticsearchClientError {
     this.name = 'ResponseError'
     if (meta.body && meta.body.error && meta.body.status) {
       if (Array.isArray(meta.body.error.root_cause)) {
-        this.message = meta.body.error.root_cause.map(entry => `[${entry.type}] Reason: ${entry.reason}`).join(' > ')
+        this.message = meta.body.error.root_cause.map(entry => `[${entry.type}] Reason: ${entry.reason}`).join('; ')
       } else {
         this.message = 'Response Error'
       }
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -91,9 +91,8 @@ class ResponseError extends ElasticsearchClientError {
     Error.captureStackTrace(this, ResponseError)
     this.name = 'ResponseError'
     if (meta.body && meta.body.error && meta.body.status) {
-      const rootCause = Array.isArray(meta.body.error.root_cause) && meta.body.error.root_cause[0]
-      if (rootCause != null) {
-        this.message = `[${rootCause.type}]. Reason: ${rootCause.reason}`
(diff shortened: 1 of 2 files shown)


lib/errors.js
```diff
@@ -90,7 +90,16 @@ class ResponseError extends ElasticsearchClientError {
     super('Response Error')
     Error.captureStackTrace(this, ResponseError)
     this.name = 'ResponseError'
-    this.message = (meta.body && meta.body.error && meta.body.error.type) || 'Response Error'
+    if (meta.body && meta.body.error && meta.body.status) {
+      const rootCause = Array.isArray(meta.body.error.root_cause) && meta.body.error.root_cause[0]
```
mshustov, line 94: Shouldn't we provide all the problems listed in `root_cause`?

delvedor: Do you mean unwrap the entire array?

mshustov: yes

delvedor: We can do that, but how would you structure the error message?
Just chain every element of the array? It can be done, but we can't have an easily predictable error message length. Would that be an issue?

mshustov: >  It can be done, but we can't have an easily predictable error message length. Would that be an issue?

I don't think it changes a lot: Elasticsearch can add a long error message even for a single reason.

mshustov (approved):

PR #1460, opened 2021-05-03
Catch HEAD errors

delvedor:
In case of an HTTP error and a `HEAD` request, the body should not be cast to boolean (unless the error is a 404).

Related: https://github.com/elastic/kibana/issues/98890

### DIFF
--- a/test/unit/transport.test.js
+++ b/test/unit/transport.test.js
@@ -1613,7 +1613,7 @@ test('Should cast to boolean HEAD request', t => {
   })
 
   t.test('4xx response', t => {
-    t.plan(2)
+    t.plan(3)
 
     const pool = new ConnectionPool({ Connection: MockConnection })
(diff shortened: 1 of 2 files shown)


test/unit/transport.test.js
```diff
@@ -1633,12 +1633,13 @@ test('Should cast to boolean HEAD request', t => {
       path: '/400'
     }, (err, { body, statusCode }) => {
       t.ok(err instanceof ResponseError)
+      t.false(typeof err.body === 'boolean')
```
mshustov, line 1636: shouldn't we test `body` contains `payload`?

mshustov (approved):

test/unit/transport.test.js
delvedor: It should never contain `payload` 🤔 
Also, if case of HEAD request, the body should be either a boolean or something else :)

PR #1487, opened 2021-07-12
Verify connection to Elasticsearch

delvedor:
Follows this logic for verifying a connection to Elasticsearch:

Before the first API request:
1. Make an API request to `/`, inspect the response:
2. If call to `/` fails with `401` or `403` pass the check and show a warning (message will be linked later). This happens if the  `monitor` permission missing for user. The subsequent checks must be ignored.
2. If there's no `version` field or if the `version` field value is <6.0.0 raise an error.
3. If there is a `version` field and it's >= 6.0.0 and <7.0.0:
    - If there is no `tagline` field or if the `tagline` field value isn't `You know, for Search` raise an error.
4. If there is a `version` field and it's >=7.0.0 and <7.14.0:
    - If there is no `tagline` field or if the `tagline` field value isn't `You know, for Search` raise an error.
    - If there is no `build_flavor` field or if the `build_flavor` field value isn't `default` raise an error.
5. If there is a `version` field and it's >=7.14.0:
    - If there is no `X-Elastic-Product` HTTP header in the response or if the `X-Elastic-Product` HTTP header value isn't `Elasticsearch` raise an error.

### DIFF
--- a/index.js
+++ b/index.js
@@ -256,8 +256,9 @@ class Client extends ESAPI {
 
     const client = new Client(options)
     // sync product check
-    const tSymbols = Object.getOwnPropertySymbols(this.transport)
-    client.transport[tSymbols[0]] = this.transport[tSymbols[0]]
+    const tSymbol = Object.getOwnPropertySymbols(this.transport)
+      .filter(symbol => symbol.description === 'product check')[0]
+    client.transport[tSymbol] = this.transport[tSymbol]
     // Add parent extensions
     if (this[kExtensions].length > 0) {
       this[kExtensions].forEach(({ name, opts, fn }) => {
--- a/index.js
+++ b/index.js
@@ -251,6 +251,9 @@ class Client extends ESAPI {
     }
 
     const client = new Client(options)
+    // sync product check
+    const tSymbols = Object.getOwnPropertySymbols(this.transport)
+    client.transport[tSymbols[0]] = this.transport[tSymbols[0]]
     // Add parent extensions
     if (this[kExtensions].length > 0) {
       this[kExtensions].forEach(({ name, opts, fn }) => {
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -538,6 +538,7 @@ class Transport {
   }
 
   productCheck () {
+    debug('Start product check')
     this[kProductCheck] = 1
     this.request({
       method: 'GET',
@@ -545,6 +546,7 @@ class Transport {
     }, (err, result) => {
       this[kProductCheck] = 3
       if (err) {
+        debug('Product check failed', err)
         if (err.statusCode === 401 || err.statusCode === 403) {
           this[kProductCheck] = 2
           process.emitWarning('The client is unable to verify that the server is Elasticsearch due to security privileges on the server side. Some functionality may not be compatible if the server is running an unsupported product.')
@@ -554,7 +556,9 @@ class Transport {
           productCheckEmitter.emit('product-check', false)
         }
       } else {
+        debug('Checking elasticsearch version', result.body, result.headers)
         if (result.body.version == null || typeof result.body.version.number !== 'string') {
+          debug('Can\'t access Elasticsearch version')
           return productCheckEmitter.emit('product-check', false)
         }
         const tagline = result.body.tagline
@@ -565,17 +569,21 @@ class Transport {
           return productCheckEmitter.emit('product-check', false)
         } else if (major >= 6 && major < 7) {
           if (tagline !== 'You Know, for Search') {
+            debug('Bad tagline')
             return productCheckEmitter.emit('product-check', false)
           }
         } else if (major === 7 && minor < 14) {
           if (tagline !== 'You Know, for Search' || result.body.version.build_flavor !== 'default') {
+            debug('Bad tagline or build_flavor')
             return productCheckEmitter.emit('product-check', false)
           }
         } else {
           if (result.headers['x-elastic-product'] !== 'Elasticsearch') {
+            debug('x-elastic-product not recognized')
             return productCheckEmitter.emit('product-check', false)
           }
         }
+        debug('Valid Elasticsearch distribution')
         this[kProductCheck] = 2
         productCheckEmitter.emit('product-check', true)
       }
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -26,12 +26,12 @@ const buffer = require('buffer')
 const ms = require('ms')
 const { EventEmitter } = require('events')
 const {
-  ElasticsearchClientError,
   ConnectionError,
   RequestAbortedError,
   NoLivingConnectionsError,
   ResponseError,
-  ConfigurationError
+  ConfigurationError,
+  ProductNotSupportedError
 } = require('./errors')
 
 const noop = () => {}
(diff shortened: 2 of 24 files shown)


mshustov:
Kibana implements a periodical version check as well https://github.com/elastic/kibana/blob/ef6b84dfa544a4677e1a33e911ef484979b5af17/src/core/server/elasticsearch/version_check/ensure_es_version.ts#L142 but both checks complement each other. 

> Before the first API request:

I guess we need to handle the incompatibility error in a special way to stop the Kibana instance. It might require additional work on the Kibana side to trigger `@elastic/elasticsearch` version check. Are you going to ship the check in v8.0 or v7.x?

lib/Transport.js
```diff
@@ -347,91 +356,124 @@ class Transport {
...
-        params.headers['content-length'] = '' + Buffer.byteLength(params.body)
-        makeRequest()
+        // the very first request triggers the product check
+        if (this[kProductCheck] === 0) {
+          this.productCheck()
+        }
       }
+    // the product check is finished and it's not Elasticsearch
+    } else if (this[kProductCheck] === 3) {
+      const err = new ElasticsearchClientError('The client noticed that the server is not Elasticsearch and we do not support this unknown product.')
```
mshustov, line 471: Maybe it makes sense to have a dedicated error class for this case? It would allow Kibana to distinguish between this critical error and another `ElasticsearchClientError`

delvedor:
@mshustov the check will be shipped in v7.

lib/Transport.js
delvedor: Agreed!

mshustov:
@delvedor in which version? I will fill out an ER for the Kibana.

test/acceptance/product-check.test.js
```diff
@@ -0,0 +1,1077 @@
...
+  const MockConnection = buildMockConnection({
+    onRequest (params) {
+      return {
+        statusCode: 200,
+        body: {
+          name: '1ef419078577',
+          cluster_name: 'docker-cluster',
+          cluster_uuid: 'cQ5pAMvRRTyEzObH4L5mTA',
+          version: {
+            number: '8.0.0-SNAPSHOT',
```
joshdover, line 104: nit: Should we be using `SNAPSHOT` in the test data? (applies to all of these tests)

test/acceptance/product-check.test.js
```diff
@@ -0,0 +1,1077 @@
...
+          version: {
+            number: '7.13.0-SNAPSHOT',
+            build_flavor: 'default',
+            build_type: 'docker',
+            build_hash: '5fb4c050958a6b0b6a70a6fb3e616d0e390eaac3',
+            build_date: '2021-07-10T01:45:02.136546168Z',
+            build_snapshot: true,
+            lucene_version: '8.9.0',
+            minimum_wire_compatibility_version: '7.15.0',
+            minimum_index_compatibility_version: '7.0.0'
```
joshdover, line 173: nit: I don't think these are correct for any of the 7.x/6.x/5.x/4.x tests. They're not being used by the check right now, but it seems worth making sure this tests data is close to being accurate in case we do use these fields in the future.

test/acceptance/product-check.test.js
```diff
@@ -0,0 +1,1077 @@
...
+    t.equal(err.message, 'The client noticed that the server is not Elasticsearch and we do not support this unknown product.')
+  })
+})
+
+test('Auth error', t => {
+  t.plan(8)
+  const MockConnection = buildMockConnection({
+    onRequest (params) {
+      return {
+        statusCode: 401,
```
joshdover, line 399: We should add test for 403 case for full coverage

lib/Transport.js
```diff
@@ -494,6 +536,51 @@ class Transport {
...
+    this[kProductCheck] = 1
+    this.request({
+      method: 'GET',
+      path: '/'
+    }, (err, result) => {
+      this[kProductCheck] = 3
+      if (err) {
+        if (err.statusCode === 401 || err.statusCode === 403) {
+          this[kProductCheck] = 2
+          process.emitWarning('The client is unable to verify that the server is Elasticsearch due to security privileges on the server side. Some functionality may not be compatible if the server is running an unsupported product.')
```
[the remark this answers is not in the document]
delvedor: I like your suggestion, but this message cannot be changed as it has already been approved by product :)

index.js
```diff
@@ -255,6 +255,9 @@ class Client extends ESAPI {
     }
 
     const client = new Client(options)
+    // sync product check
+    const tSymbols = Object.getOwnPropertySymbols(this.transport)
+    client.transport[tSymbols[0]] = this.transport[tSymbols[0]]
```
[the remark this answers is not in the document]
delvedor: Good call out, I'll see what I can do.

lib/Transport.js
```diff
@@ -494,6 +536,51 @@ class Transport {
...
+          productCheckEmitter.emit('product-check', true)
+        } else {
+          this[kProductCheck] = 0
+          productCheckEmitter.emit('product-check', false)
+        }
+      } else {
+        if (result.body.version == null || typeof result.body.version.number !== 'string') {
```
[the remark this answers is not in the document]
delvedor: As far as I know, when you are checking for both `null` and `undefined`, doing `== null` does the job pretty well, as it's testing for both.

test/acceptance/product-check.test.js
delvedor: Given that we are only looking for the first two numbers it's fine :)

test/acceptance/product-check.test.js
delvedor: You are right, I only changed the fields that we need to validate.
Given that from 7.14 we'll only test the response headers there is no need to make this too much future proof. I'll revisit if needed :)

joshdover (approved):

lib/Transport.js
```diff
@@ -494,6 +536,59 @@ class Transport {
...
+        const minor = Number(version[1])
+        if (major < 6) {
+          return productCheckEmitter.emit('product-check', false)
+        } else if (major >= 6 && major < 7) {
+          if (tagline !== 'You Know, for Search') {
+            debug('Bad tagline')
+            return productCheckEmitter.emit('product-check', false)
+          }
+        } else if (major === 7 && minor < 14) {
+          if (tagline !== 'You Know, for Search' || result.body.version.build_flavor !== 'default') {
```
marcbachmann, line 576: FYI this breaks support for the oss docker images <= 7.10.1.
The build_flavor in there s `oss`.
```
{
  "name" : "elasticsearch-production-2",
  "cluster_name" : "elasticsearch-bmg-prd",
  "cluster_uuid" : "41tljEp7T66B-13N2zaE4A",
  "version" : {
    "number" : "7.10.1",
    "build_flavor" : "oss",
    "build_type" : "docker",
    "build_hash" : "1c34507e66d7db1211f66f3513706fdf548736aa",
    "build_date" : "2020-12-05T01:00:33.671820Z",
    "build_snapshot" : false,
    "lucene_version" : "8.7.0",
    "minimum_wire_compatibility_version" : "6.8.0",
    "minimum_index_compatibility_version" : "6.0.0-beta1"
  },
  "tagline" : "You Know, for Search"
}
```

While an upgrade to a newer image fixes the incompatibility, you might want to adopt the check in here.

delvedor: Hello! Please see https://github.com/elastic/elasticsearch-js/issues/1519.

marcbachmann: Thanks Thomas, that's then on purpose as I've assumed. There's just no mention in this PR, which could have the change a bit more transparent.

PR #1499, opened 2021-07-22
Support CA fingerprint validation

delvedor:
As titled.

Usage: 
```js
const client = new Client({
  node: 'https://localhost:9200',
  caFingerprint: 'string'
})
```

TODO:
- [x] More unit test
- [x] Docs
- [ ] ~~Integration test with Elasticsearch~~

### DIFF
--- a/docs/basic-config.asciidoc
+++ b/docs/basic-config.asciidoc
@@ -255,4 +255,8 @@ const client = new Client({
 |`boolean`, `'proto'`, `'constructor'` - By the default the client will protect you against prototype poisoning attacks. Read https://web.archive.org/web/20200319091159/https://hueniverse.com/square-brackets-are-the-enemy-ff5b9fd8a3e8?gi=184a27ee2a08[this article] to learn more. If needed you can disable prototype poisoning protection entirely or one of the two checks. Read the `secure-json-parse` https://github.com/fastify/secure-json-parse[documentation] to learn more. +
 _Default:_ `false`
 
+|`caFingerprint`
+|`string` - If configured, verify the supplied certificate authority fingerprint. +
+_Default:_ `null`
+
 |===
--- a/docs/basic-config.asciidoc
+++ b/docs/basic-config.asciidoc
@@ -256,7 +256,7 @@ const client = new Client({
 _Default:_ `false`
 
 |`caFingerprint`
-|`string` - If configured, verify the supplied CA certificate fingerprint. You must configure a SHA256 diagest. +
+|`string` - If configured, verify that the fingerprint of the CA certificate that has signed the certificate of the server matches the supplied fingerprint. Only accepts SHA256 digest fingerprints. +
 _Default:_ `null`
 
 |===
--- a/docs/basic-config.asciidoc
+++ b/docs/basic-config.asciidoc
@@ -256,7 +256,7 @@ const client = new Client({
 _Default:_ `false`
 
 |`caFingerprint`
-|`string` - If configured, verify the supplied certificate authority fingerprint. +
+|`string` - If configured, verify the supplied CA certificate fingerprint. +
 _Default:_ `null`
 
 |===
@@ -280,4 +280,3 @@ const client = new Client({
   disablePrototypePoisoningProtection: true
 })
 ----
-
--- a/docs/basic-config.asciidoc
+++ b/docs/basic-config.asciidoc
@@ -256,7 +256,7 @@ const client = new Client({
 _Default:_ `false`
 
 |`caFingerprint`
-|`string` - If configured, verify the supplied CA certificate fingerprint. +
+|`string` - If configured, verify the supplied CA certificate fingerprint. You must configure a SHA256 diagest. +
 _Default:_ `null`
 
 |===
--- a/docs/connecting.asciidoc
+++ b/docs/connecting.asciidoc
@@ -180,7 +180,7 @@ const client = new Client({
 [[auth-ca-fingerprint]]
 ==== CA fingerprint
 
-To improve the security of your connection to Elasticsearch, you can provide
+You can configure <not sure how we refer to the client in docs here> to only trust  certificates that are signed by a specific CA certificate ( CA certificate pinning ) by providing a `caFingerprint` option. This will verify that the fingerprint of the CA certificate that has signed the certificate of the server matches the supplied value.
 a `caFingerprint` option, which will verify the supplied certificate authority fingerprint.
 You must configure a SHA256 digest.
 
--- a/docs/connecting.asciidoc
+++ b/docs/connecting.asciidoc
@@ -176,6 +176,28 @@ const client = new Client({
 })
 ----
 
+[discrete]
+[[auth-ca-fingerprint]]
+==== CA fingerprint
+
+To improve the security of your connection to Elasticsearch, you can provide
+a `caFingerprint` option, which will verify the supplied certificate authority fingerprint.
+
+[source,js]
+----
+const { Client } = require('@elastic/elasticsearch')
+const client = new Client({
+  node: 'https://example.com'
+  auth: { ... },
+  // the fingerprint (SHA256) of the CA certificate that is used to sign the certificate that the Elasticsearch node presents for TLS.
+  caFingerprint: 'SHA-256-DER',
+  ssl: {
+    // might be required if it's a self-signed certificate
+    rejectUnauthorized: false
+  }
+})
+----
+
 
 [discrete]
 [[client-usage]]
--- a/docs/connecting.asciidoc
+++ b/docs/connecting.asciidoc
@@ -180,7 +180,7 @@ const client = new Client({
 [[auth-ca-fingerprint]]
 ==== CA fingerprint
 
-You can configure <not sure how we refer to the client in docs here> to only trust  certificates that are signed by a specific CA certificate ( CA certificate pinning ) by providing a `caFingerprint` option. This will verify that the fingerprint of the CA certificate that has signed the certificate of the server matches the supplied value.
+You can configure the client to only trust certificates that are signed by a specific CA certificate ( CA certificate pinning ) by providing a `caFingerprint` option. This will verify that the fingerprint of the CA certificate that has signed the certificate of the server matches the supplied value.
 a `caFingerprint` option, which will verify the supplied certificate authority fingerprint.
 You must configure a SHA256 digest.
 
--- a/docs/connecting.asciidoc
+++ b/docs/connecting.asciidoc
@@ -182,6 +182,7 @@ const client = new Client({
 
 To improve the security of your connection to Elasticsearch, you can provide
 a `caFingerprint` option, which will verify the supplied certificate authority fingerprint.
+You must configure a SHA256 digest.
 
 [source,js]
 ----
--- a/docs/connecting.asciidoc
+++ b/docs/connecting.asciidoc
@@ -190,7 +190,7 @@ const client = new Client({
   node: 'https://example.com'
   auth: { ... },
   // the fingerprint (SHA256) of the CA certificate that is used to sign the certificate that the Elasticsearch node presents for TLS.
-  caFingerprint: 'SHA-256-DER',
+  caFingerprint: '20:0D:CA:FA:76:...',
   ssl: {
     // might be required if it's a self-signed certificate
     rejectUnauthorized: false
--- a/index.js
+++ b/index.js
@@ -323,12 +323,7 @@ function getAuth (node) {
 
 function isHttpConnection (node) {
   if (Array.isArray(node)) {
-    for (const n of node) {
-      if (new URL(n).protocol === 'http:') {
-        return true
-      }
-    }
-    return false
+    return node.some((n) => new URL(n).protocol === 'http:')
   } else {
     return new URL(node).protocol === 'http:'
   }
--- a/index.js
+++ b/index.js
@@ -102,6 +102,7 @@ class Client extends ESAPI {
         suggestCompression: false,
         compression: false,
         ssl: null,
+        certFingerprint: null,
         agent: null,
         headers: {},
         nodeFilter: null,
@@ -146,6 +147,7 @@ class Client extends ESAPI {
         Connection: options.Connection,
         auth: options.auth,
         emit: this[kEventEmitter].emit.bind(this[kEventEmitter]),
+        certFingerprint: options.certFingerprint,
         sniffEnabled: options.sniffInterval !== false ||
                       options.sniffOnStart !== false ||
                       options.sniffOnConnectionFault !== false
--- a/index.js
+++ b/index.js
@@ -102,7 +102,7 @@ class Client extends ESAPI {
         suggestCompression: false,
         compression: false,
         ssl: null,
-        certFingerprint: null,
+        caFingerprint: null,
         agent: null,
         headers: {},
         nodeFilter: null,
@@ -147,7 +147,7 @@ class Client extends ESAPI {
         Connection: options.Connection,
         auth: options.auth,
         emit: this[kEventEmitter].emit.bind(this[kEventEmitter]),
-        certFingerprint: options.certFingerprint,
+        caFingerprint: options.caFingerprint,
         sniffEnabled: options.sniffInterval !== false ||
                       options.sniffOnStart !== false ||
                       options.sniffOnConnectionFault !== false
--- a/index.js
+++ b/index.js
@@ -117,6 +117,10 @@ class Client extends ESAPI {
         disablePrototypePoisoningProtection: false
       }, opts)
 
+    if (options.caFingerprint !== null && isHttpConnection(opts.node || opts.nodes)) {
+      throw new ConfigurationError('You can\'t configure the caFingerprint with a http connection')
+    }
+
     if (process.env.ELASTIC_CLIENT_APIVERSIONING === 'true') {
       options.headers = Object.assign({ accept: 'application/vnd.elasticsearch+json; compatible-with=7' }, options.headers)
     }
@@ -314,6 +318,19 @@ function getAuth (node) {
         password: decodeURIComponent(node.url.password)
       }
     }
+  }
+}
+
+function isHttpConnection (node) {
+  if (Array.isArray(node)) {
+    for (const n of node) {
+      if (new URL(n).protocol === 'http:') {
+        return true
+      }
+    }
+    return false
+  } else {
+    return new URL(node).protocol === 'http:'
   }
 }
 
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -139,7 +139,7 @@ class Connection {
           // Check if fingerprint matches
           /* istanbul ignore else */
           if (this.caFingerprint !== issuerCertificate.fingerprint256) {
-            onError(new Error('Server certificate CA fingerprint does not match the value configured in  caFingerprint'))
+            onError(new Error('Server certificate CA fingerprint does not match the value configured in caFingerprint'))
             request.once('error', () => {}) // we need to catch the request aborted error
             return request.abort()
           }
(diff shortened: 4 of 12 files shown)


lib/Connection.js
```diff
@@ -123,10 +124,34 @@ class Connection {
...
+    const onSocket = socket => {
+      /* istanbul ignore else */
+      if (!socket.isSessionReused()) {
+        socket.once('secureConnect', () => {
+          // Check if certificate is valid
+          if (socket.authorized === false) {
+            onError(new Error(socket.authorizationError))
+            request.once('error', () => {}) // we need to catch the request aborted error
+            return request.abort()
+          }
```
delvedor, line 136: I’m not entirely sure this is needed, as this check would catch the self-signed certificate.
One of the test is failing because of this check, if we remove it, everything works as expected.

azasypkin: I suspect it's not needed since we're going to validate a fingerprint of the top-most CA certificate, that will most-likely be a self-signed one (from the tech spec `The fingerprint (SHA256) of the CA certificate that is used to sign the certificate that the Elasticsearch node presents for TLS. `). @jkakavas is my understanding correct?

lib/Connection.js
```diff
@@ -123,10 +124,34 @@ class Connection {
...
+
+          // Check if fingerprint matches
+          if (this.certFingerprint !== socket.getPeerCertificate().fingerprint) {
```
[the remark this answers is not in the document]
azasypkin: __issue:__ I believe `socket.getPeerCertificate()` returns a leaf/node certificate only, but we need to retrieve a full chain to get fingerprint for the CA certificate. If so, we'd need to use `request.socket.getPeerCertificate(true)` and rely on `issuerCertificate` property to recursively bubble-up to the top-most one? Somewhat similar to what we do for [PKI in Kibana](https://github.com/elastic/kibana/blob/80876620f94d4e17ce69e6ac1f578cc484b8c89b/x-pack/plugins/security/server/authentication/providers/pki.ts#L317)

thomheymann (commented):
I haven't tested this locally yet but added a few comments based on the code. 

One general question I have is how we can avoid the "self signed certificate in certificate chain" error nodejs will throw during interactive setup when connecting to the (at that point) untrusted cert. We can't ask users to set NODE_TLS_REJECT_UNAUTHORIZED=0 and setting `ssl.verificationMode=none` in the elasticsearch client doesn't seem to prevent it.

lib/Connection.js
```diff
@@ -123,10 +124,34 @@ class Connection {
       callback(new RequestAbortedError(), null)
     }
 
+    const onSocket = socket => {
+      /* istanbul ignore else */
+      if (!socket.isSessionReused()) {
```
azasypkin, line 129: __question:__ just for my own education, if TLS session is re-used, NodeJS doesn't issue `secureConnect`?

test/utils/buildServer.js
```diff
@@ -35,6 +36,13 @@ const secureOpts = {
   cert: readFileSync(join(__dirname, '..', 'fixtures', 'https.cert'), 'utf8')
 }
 
+const certFingerprint = getFingerprint(secureOpts.cert
+  .split('\n')
+  .slice(1, -1)
```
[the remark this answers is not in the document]
delvedor: Given this is just for unit test purposes I'm fine leaving it as is.
I'll work on a more advanced integration test with Elasticsearch.

lib/Connection.js
```diff
@@ -42,6 +42,7 @@ class Connection {
     this.headers = prepareHeaders(opts.headers, opts.auth)
     this.deadCount = 0
     this.resurrectTimeout = 0
+    this.certFingerprint = opts.certFingerprint
```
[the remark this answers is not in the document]
delvedor: Good call, I would say that `caFingerprint` is the best, so we avoid the double "certificate" repetition :)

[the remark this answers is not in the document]
azasypkin: Technically we don't repeat since it's `Certificate Authority certificate fingerprint`, but either works for me :slightly_smiling_face:

lib/Connection.js
delvedor: It might be, but this check avoids the allocation of yet another event emitter with its own closure :)

docs/connecting.asciidoc
```diff
@@ -176,6 +176,28 @@ const client = new Client({
...
+a `caFingerprint` option, which will verify the supplied certificate authority fingerprint.
+
+[source,js]
+----
+const { Client } = require('@elastic/elasticsearch')
+const client = new Client({
+  node: 'https://example.com'
+  auth: { ... },
+  // the fingerprint (SHA256) of the CA certificate that is used to sign the certificate that the Elasticsearch node presents for TLS.
+  caFingerprint: 'SHA-256-DER',
```
sethmlarson, line 193: This example might be better if it had some example hex values? Without it might be tough to figure out what to put here?

ie `'20:0D:CA:FA:76:...'`

sethmlarson (approved):
LGTM!

thomheymann (commented):
I've tested this locally and the fingerprint is correctly verified. 🥳

However, this only works when running node with `NODE_TLS_REJECT_UNAUTHORIZED=0` flag. Otherwise we don't get to the verification step since node throws an error for self signed certificates. 

Asking users to run Kibana with that flag is not a viable option. Appreciate that this is node throwing an error but could we set rejectUnauthorized to false when `caFingerprint` option is set on the client to avoid this? Is there another option you can think of? 

```
$ node test.js

ERROR ConnectionError: self signed certificate in certificate chain
    at ClientRequest.onError (/Users/thom/Repositories/elasticsearch-js/lib/Connection.js:116:16)
    at ClientRequest.emit (events.js:375:28)
    at TLSSocket.socketErrorListener (_http_client.js:475:9)
    at TLSSocket.emit (events.js:375:28)
    at emitErrorNT (internal/streams/destroy.js:106:8)
    at emitErrorCloseNT (internal/streams/destroy.js:74:3)
    at processTicksAndRejections (internal/process/task_queues.js:82:21) 
```

delvedor:
> Appreciate that this is node throwing an error but could we set rejectUnauthorized to false when caFingerprint option is set on the client to avoid this? Is there another option you can think of?

I'm fine to setting `rejectUnauthorized` by default when `caFingerprint` is configured, I'll do it :)

thomheymann (approved):
As discussed offline, let's leave this as opt in and let the consumer decide whether to disable `rejectUnauthorized` or not for now. 

Otherwise LGTM! Thanks for implementing this so swiftly. 🙏

docs/basic-config.asciidoc
```diff
@@ -255,6 +255,10 @@ const client = new Client({
 |`boolean`, `'proto'`, `'constructor'` - By the default the client will protect you against prototype poisoning attacks. Read https://web.archive.org/web/20200319091159/https://hueniverse.com/square-brackets-are-the-enemy-ff5b9fd8a3e8?gi=184a27ee2a08[this article] to learn more. If needed you can disable prototype poisoning protection entirely or one of the two checks. Read the `secure-json-parse` https://github.com/fastify/secure-json-parse[documentation] to learn more. +
 _Default:_ `false`
 
+|`caFingerprint`
+|`string` - If configured, verify the supplied certificate authority fingerprint. +
```
azasypkin, line 259: __optional nit:__ maybe that's just me, but it sounds a bit confusing, and I'd rather say (since CA is a body/organization that doesn't have a fingerprint :slightly_smiling_face: ):
```suggestion
|`string` - If configured, verify the supplied CA certificate fingerprint. +
```
It has become very common to call CA certificate as just CA though, so up you you.

azasypkin (approved):
LGTM from Kibana standpoint, thanks! Just one issue we discussed over Slack: handle/test the case when `caFingerprint` is provided, but ES client tries to connect to HTTP address.

docs/basic-config.asciidoc
thomheymann: I would also add a sentence that the fingerprint is SHA-256 digest in HEX format, maybe with an example

test/unit/client.test.js
```diff
@@ -1498,3 +1499,88 @@ test('Bearer auth', t => {
...
+    t.equal(err.message, 'You can\'t configure the caFingerprint with a http connection')
+  }
+})
+
+test('caFingerprint can\'t be configured over http / 2', t => {
+  t.plan(2)
+
+  try {
+    new Client({ // eslint-disable-line
+      nodes: ['http://localhost:9200'],
```
thomheymann, line 1564: What's the behaviour when someone adds multiple nodes with different protocols?

`['http://localhost:9200', 'https://localhost:9201']`

delvedor: The first one that fails causes the exception.

index.js
```diff
@@ -315,6 +321,19 @@ function getAuth (node) {
...
 }
 
+function isHttpConnection (node) {
+  if (Array.isArray(node)) {
+    for (const n of node) {
+      if (new URL(n).protocol === 'http:') {
+        return true
+      }
+    }
+    return false
```
azasypkin, line 331: __optional nit:__ (a bit easier to read, but it's more of a personal preference)
```suggestion
    return node.some((n) => new URL(n).protocol === 'http:')
```

docs/basic-config.asciidoc
```diff
@@ -255,6 +255,10 @@ const client = new Client({
 |`boolean`, `'proto'`, `'constructor'` - By the default the client will protect you against prototype poisoning attacks. Read https://web.archive.org/web/20200319091159/https://hueniverse.com/square-brackets-are-the-enemy-ff5b9fd8a3e8?gi=184a27ee2a08[this article] to learn more. If needed you can disable prototype poisoning protection entirely or one of the two checks. Read the `secure-json-parse` https://github.com/fastify/secure-json-parse[documentation] to learn more. +
 _Default:_ `false`
 
+|`caFingerprint`
```
azasypkin, line 258: __nit:__ I agree with @thomheymann that it'd make sense to mention that we're expecting SHA-256 digest - otherwise consumer may not realize why their SHA-1 fingerprint doesn't work.

We could also easily check if provided SHA-1 or SHA-256 is valid (based on the predetermined string length and knowing that it's a HEX string) and throw if it doesn't pass validation, but I'll leave to you :slightly_smiling_face:

azasypkin (approved):
Still LGTM!

lib/Connection.js
jkakavas: > @jkakavas is my understanding correct?

Yes. We elected to use the CA certificate for the pinning as it originally felt like it provided for more options. It _will_ be a self signed CA certificate that Elasticsearch generates on startup

jkakavas:
> As discussed offline, let's leave this as opt in and let the consumer decide whether to disable rejectUnauthorized or not for now.

Just validating my understanding that Kibana will set this so that the users won't need to configure something extra. Can this be set per connection and only set it for the initial call to the enroll API ? We wouldn't want to set this globally off for Kibana

lib/Connection.js
```diff
@@ -123,10 +124,28 @@ class Connection {
...
     }
 
+    const onSocket = socket => {
+      /* istanbul ignore else */
+      if (!socket.isSessionReused()) {
+        socket.once('secureConnect', () => {
+          // Check if fingerprint matches
+          /* istanbul ignore else */
+          if (this.caFingerprint !== getIssuerCertificate(socket).fingerprint256) {
+            onError(new Error('Fingerprint does not match'))
```
jkakavas, line 134:
```suggestion
            onError(new Error('Server certificate CA fingerprint does not match the value configured in  caFingerprint'))
```

Not sure how detailed we want to be in our error messages in the client but I found the above to be a bit cryptic

jkakavas (commented):
Added just a few comments, thanks for tackling this so quickly!

index.js
```diff
@@ -116,6 +117,10 @@ class Client extends ESAPI {
         disablePrototypePoisoningProtection: false
       }, opts)
 
+    if (options.caFingerprint !== null && isHttpConnection(opts.node || opts.nodes)) {
```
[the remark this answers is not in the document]
delvedor: We don't :) It has never been an issue, so for now, I won't add an additional check for it.

thomheymann:
> > As discussed offline, let's leave this as opt in and let the consumer decide whether to disable rejectUnauthorized or not for now.
> 
> Just validating my understanding that Kibana will set this so that the users won't need to configure something extra. Can this be set per connection and only set it for the initial call to the enroll API ? We wouldn't want to set this globally off for Kibana

Correct, Kibana will disable `rejectUnauthorized` for the enrolment call so that we can connect to Elasticsearch until we have established trust. 

All other requests to Elasticsearch will keep that security check enabled.

index.js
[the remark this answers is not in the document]
jkakavas: I was just curious

lib/Connection.js
```diff
@@ -340,5 +368,34 @@ function prepareHeaders (headers = {}, auth) {
...
+    if (certificate.subject.C === certificate.issuer.C &&
+        certificate.subject.ST === certificate.issuer.ST &&
+        certificate.subject.L === certificate.issuer.L &&
+        certificate.subject.O === certificate.issuer.O &&
+        certificate.subject.OU === certificate.issuer.OU &&
+        certificate.subject.CN === certificate.issuer.CN) {
+      break
+    }
+
+    // For self-signed certificates, `issuerCertificate` may be a circular reference.
```
jkakavas, line 389: We probably never reach this as by definition a self-signed certificate has an issuer that equals its subject. Practically, the fingerprint256 is the sha256 hash of the contents of the certificate so if the previous `if` clause is not true, there is no way for this `if` clause to be true ( in order for us to reach here, the certificate subject is different than the issuerCertificate subject so the fingerprint can't match) . I say we skip this entirely

docs/connecting.asciidoc
```diff
@@ -176,6 +176,29 @@ const client = new Client({
 })
 ----
 
+[discrete]
+[[auth-ca-fingerprint]]
+==== CA fingerprint
+
+You can configure <not sure how we refer to the client in docs here> to only trust  certificates that are signed by a specific CA certificate ( CA certificate pinning ) by providing a `caFingerprint` option. This will verify that the fingerprint of the CA certificate that has signed the certificate of the server matches the supplied value.
```
jkakavas, line 183:
```suggestion
You can configure the client to only trust certificates that are signed by a specific CA certificate ( CA certificate pinning ) by providing a `caFingerprint` option. This will verify that the fingerprint of the CA certificate that has signed the certificate of the server matches the supplied value.
```

test/unit/client.test.js
```diff
@@ -1498,3 +1499,88 @@ test('Bearer auth', t => {
...
+      caFingerprint: 'FO:OB:AR'
+    })
+    t.fail('shuld throw')
```
jkakavas, line 1567:
```suggestion
    t.fail('should throw')
```

jkakavas (approved):
LGTM with my latest comments addressed, I don't think this needs another round. Thanks Tomas!

lib/Connection.js
delvedor: Good call, it would be even better to just compare the two fingerprints then, which is slightly faster than checking every single property.

jkakavas: I think that's fine yes. I'm a bit worried by the nodejs documentation that says "_may be_ a circular reference" because I would expect this this to _always_ be a circular reference for self signed certs and this is what we're checking here, but this worry would be the same for the check we removed, so ++ for the change, let's see if we come about any issues

lib/Connection.js
```diff
@@ -340,5 +360,24 @@ function prepareHeaders (headers = {}, auth) {
...
+    if (certificate.issuerCertificate !== undefined) {
+      // For self-signed certificates, `issuerCertificate` may be a circular reference.
+      /* istanbul ignore else */
+      if (certificate.fingerprint256 === certificate.issuerCertificate.fingerprint256) {
+        break
+      }
+      /* istanbul ignore next */
+      certificate = certificate.issuerCertificate
+    } else {
+      break
```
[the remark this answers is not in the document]
azasypkin: > when issuerCertificate is undefined, the certificate is invalid or malformed, let's not try to use it.

Hmm, it seems that missing `issuerCertificate` in certificate is one thing and missing `issuerCertificate` property in a certificate parsed by NodeJS is a different thing. If host just doesn't return a full chain, I _guess_, we'll have an empty `issuerCertificate` property (or if we pass `false` to `socket.getPeerCertificate`).

If ES is configured to not return full chain do we treat it as a misconfiguration @jkakavas or you meant something else?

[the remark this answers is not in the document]
jkakavas: >  missing issuerCertificate in certificate is one thing and missing issuerCertificate property in a certificate parsed by NodeJS is a different thing

Correct. I focused on the former I guess :/ Still though, we're in a method called `getIssuerCertificate` and we shouldn't return a certificate just because we couldn't get it's issuer. If ES doesn't return the full chain, then `caFingerprint` does not make much sense to be set.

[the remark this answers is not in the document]
azasypkin: > Still though, we're in a method called getIssuerCertificate and we shouldn't return a certificate just because we couldn't get it's issuer.

Yep

>  If ES doesn't return the full chain, then `caFingerprint` does not make much sense to be set.

Would it make sense to rename it to `rootCAFingerprint` then? Because we may still have a number intermediate CA certificates in the chain.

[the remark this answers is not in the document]
jkakavas: > Would it make sense to rename it to rootCAFingerprint then? Because we may still have a number intermediate CA certificates in the chain even though root CA certificate isn't there.

Possibly, it makes it clearer I think but we can handle this with documentation also. 

> And just for my own education, if I issue a Let's Encrypt certificate to use with my ES instance, I still have to configure ES with the full chain, even though root CA certificate is issued by the well-known public CA, right?

Well, we don't support the enrollment process for arbitrary configurations but if you'd want to use let's encrypt and use `caFingerprint` in a way in your es client, then yes ,you'd need to do that. 

Generically, it would make more sense for a client to support leaf certificate pinning ( as in I want to trust _this_ server certificate only ). The idea behind us using CA certificate pinning for the enrollment process is that it allows you to get a single enrollment token that might contain many addresses for all ES nodes in a cluster and you can try/selct any of them and validate the connection with a single fingerprint ( as opposed to one per node )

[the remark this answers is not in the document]
azasypkin: Yeah, but I had an impression that `caFingerprint` we've introduced here is not specific to enrollment process otherwise we should have named it differently I guess.

> Generically, it would make more sense for a client to support leaf certificate pinning ( as in I want to trust this server certificate only ).

That's actually what I was alluding to initially, pinning to the leaf (this specific server) or intermediate CA (this specific group of servers) seems to be a perfectly valid use case. My concern was that ES seems to not explicitly require full chains and moreover works fine with non-complete chains. That'd be enough for the aforementioned use case, but it won't work if we treat certificate with null/undefined `issuerCertificate` as invalid (the one that NodeJS parses, aka `if (certificate.issuerCertificate == null) return null`). 

Anyway, PR is merged, so let's if it ever becomes a problem :man_shrugging:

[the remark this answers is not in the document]
jkakavas: > That's actually what I was alluding to initially, pinning to the leaf (this specific server) or intermediate CA (this specific group of servers) seems to be a perfectly valid use case. 

Good point, I'd be :+1: This would cover security on by default enrollment process, and more generic cases at the same time. Maybe we can do in a followup ?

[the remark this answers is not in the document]
azasypkin: > Maybe we can do in a followup ?

Yeah, I'd :+1: too.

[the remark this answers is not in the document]
alimoezzi: I'm exactly having this issue. Returned certificate only has `issuer` not `issuerCertificate`. It results in `error: Unhandled Rejection at: Promise [object Promise] reason ConnectionError: Invalid or malformed certificate`. This is in contrast with Python client where `ssl_assert_fingerprint` parameter works as expected with self-signed certificate that elastic generates by default.

PR #1515, opened 2021-08-11
Add warning.name to product check security exception

### LINKED ISSUE #1510: Emitting a `process.emitWarning` kills the parent application
The following piece of code emits a warning at the process level.

https://github.com/elastic/elasticsearch-js/blob/2d1505eb2b7ad05387ca8ba8a6874f580a126ecd/lib/Transport.js#L562

Node.js is terminating the application because of that process-level warning. IMO, given that we are marking the Product Check as successful (`this[kProductCheck] = 2` and `productCheckEmitter.emit('product-check', true)`), I would expect the warning to be less destructive.

What do you think?

delvedor:
As titled.

Closes: #1510

### DIFF
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -559,7 +559,10 @@ class Transport {
         debug('Product check failed', err)
         if (err.statusCode === 401 || err.statusCode === 403) {
           this[kProductCheck] = 2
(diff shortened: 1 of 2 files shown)


afharo (approved):
LGTM! Thank you for applying this change! 🧡

PR #1526, opened 2021-08-20
Fix isHttpConnection check

delvedor:
As titled.


azasypkin:
__ACK:__ reviewing....

azasypkin (approved):
LGTM, tested locally with Kibana - everything works as expected, thanks for the quick turnaround here!

Issue #1516, opened 2021-08-11
Cannot read property 'method' of null since 7.14.0


simllll:
## 🐛 Bug Report

We suddenly get a lot of errors from elastic module.

errorHandler[500]: {"domainThrown":true,"status":500} TypeError: Cannot read property 'method' of null
2021-08-11 13:08:32at EventEmitter.<anonymous> (/app/node_modules/elastic-apm-node/lib/instrumentation/modules/@elastic/elasticsearch.js:47:52)

```
500 Error: error executing elastic query: TypeError: Cannot read property 'method' of null
2021-08-11 13:08:36at ElasticService.search (/app/node_modules/@hokify/elastic-service/src/ElasticService.ts:293:10)
2021-08-11 13:08:36at ElasticOfferQuery.execQuery (/app/node_modules/@hokify/search/src/elastic/query/ElasticGenericQuery.ts:318:44)
2021-08-11 13:08:36at processTicksAndRejections (internal/process/task_queues.js:95:5)
2021-08-11 13:08:36at ElasticOfferQuery.exec (/app/node_modules/@hokify/search/src/elastic/ElasticOfferQuery.ts:237:18)
2021-08-11 13:08:36at OfferSearchService.executeCachedQuery (/app/node_modules/@hokify/search/src/OfferSearchService.ts:707:18)
2021-08-11 13:08:36at runMethod (/app/node_modules/@hokify/node-ts-cache/src/decorator/cache.decorator.ts:37:21)
2021-08-11 13:08:36at /app/node_modules/@hokify/node-ts-cache/src/decorator/cache.decorator.ts:65:28
2021-08-11 13:08:36at OfferSearchService.doSearch (/app/node_modules/@hokify/search/src/OfferSearchService.ts:539:23)
2021-08-11 13:08:36at JobSeekerSearchOfferService.doSearch (/app/node_modules/@hokify/user-service/src/JobSeekerSearchOfferService.ts:662:25)
2021-08-11 13:08:36at executeSearch (/app/node_modules/@hokify/user-service/src/JobSeekerSearchOfferService.ts:145:42)
2021-08-11 13:08:36at JobSeekerSearchOfferService.searchOffers (/app/node_modules/@hokify/user-service/src/JobSeekerSearchOfferService.ts:180:72)
2021-08-11 13:08:36at JobSeekerSearchController.getSuggestions (/app/src/controller/jobseeker/SearchController.ts:353:7)
[...]

PR #1531, opened 2021-08-23
Always display request params and options in request event

delvedor:
As titled.

Closes: https://github.com/elastic/elasticsearch-js/issues/1516

### DIFF
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -420,8 +420,8 @@ class Transport {
       // handles request timeout
       params.timeout = toMs(options.requestTimeout || this.requestTimeout)
       if (options.asStream === true) params.asStream = true
-      meta.request.params = params
-      meta.request.options = options
+      // meta.request.params = params
+      // meta.request.options = options
 
       // handle compression
       if (params.body !== '' && params.body != null) {
@@ -452,6 +452,8 @@ class Transport {
       }
     }
 
+    meta.request.params = params
+    meta.request.options = options
     // still need to check the product or waiting for the check to finish
     if (this[kProductCheck] === 0 || this[kProductCheck] === 1) {
       // let pass info requests
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -420,8 +420,6 @@ class Transport {
       // handles request timeout
       params.timeout = toMs(options.requestTimeout || this.requestTimeout)
       if (options.asStream === true) params.asStream = true
-      // meta.request.params = params
-      // meta.request.options = options
 
       // handle compression
       if (params.body !== '' && params.body != null) {
--- a/test/acceptance/product-check.test.js
+++ b/test/acceptance/product-check.test.js
@@ -1247,3 +1247,41 @@ test('No multiple checks with child clients', t => {
     })
   }, 100)
 })
+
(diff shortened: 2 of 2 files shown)


trentm (approved):
Confirmed this fixes the crash issue in #1516 (with APM instrumentation v3.19.0 of the ES client making a request to non-ES).

### testing notes

Running a mock ES that responds like this:

```
HTTP/1.1 200 OK
X-elastic-product: Elasticsearch
content-type: application/json
Date: Mon, 23 Aug 2021 19:13:08 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Transfer-Encoding: chunked

{
  "hi": "there"
}
```

I first confirmed this hits #1516 with v3.19.0 of the APM agent (before the fix) and v7.14.0 of the ES client. "esquery.js" is a small script that uses APM and calls `esClient.search(...)`:

```
% npm ls @elastic/elasticsearch
elastic-apm-node@3.19.0 /Users/trentm/el/apm-agent-nodejs8
└── @elastic/elasticsearch@7.14.0

% node esquery.js
/Users/trentm/el/apm-agent-nodejs8/lib/instrumentation/modules/@elastic/elasticsearch.js:47
          { id: request.id, method: request.params.method, path: request.params.path })
                                                   ^

TypeError: Cannot read property 'method' of null
    at EventEmitter.<anonymous> (/Users/trentm/el/apm-agent-nodejs8/lib/instrumentation/modules/@elastic/elasticsearch.js:47:52)
    at EventEmitter.emit (events.js:314:20)
    at EventEmitter.<anonymous> (/Users/trentm/el/apm-agent-nodejs8/node_modules/@elastic/elasticsearch/lib/Transport.js:464:18)
```


Then I installed this PR's fix and ran again:

```
% npm install github:elastic/elasticsearch-js#fix-1516

% node esquery.js
{"log.level":"info","@timestamp":"2021-08-23T19:16:32.488Z","log":{"logger":"elastic-apm-node"},"ecs":{"version":"1.6.0"},"message":"Sending error to Elastic APM: {\"id\":\"71a2818c6c0a676015124ef32d8d7bc5\"}"}
```

This results in a sufficient APM trace:

- the "Elasticsearch: GET /" span
- an APM error event:

```json
{
    "error": {
        "id": "58231373ba7e0f1846342238b5bd9158",
        "timestamp": 1629746450856000,
        "parent_id": "c3c625cbf083c78b",
        "trace_id": "d11ca3e94e7ac7778b117129eff1f152",
        "transaction_id": "185a6154d541781a",
        "exception": {
            "message": "The client noticed that the server is not Elasticsearch and we do not support this unknown product.",
            "type": "ProductNotSupportedError",
            "handled": true,
            "attributes": {
                "name": "ProductNotSupportedError"
            },
            "module": "@elastic/elasticsearch",
...
```


Note that I only say "sufficient", because I think the APM instrumentation could possibly be improved here. A span is not created for the attempted "Elasticsearch: POST /_search" request. The only span is for the "Elasticsearch: GET /" product check request... and *that* request succeeded, so the outcome of the span and transaction is "success". I think ultimately we want the outcome to be "failure" for at least a span. This is independent of this PR, however. I'll open a separate ticket on the APM agent for this.

Issue #1517, opened 2021-08-11
client.on("response", ...) event does not include err=RequestAbortedError if aborted during `productCheck()`


trentm:
## 🐛 Bug Report

In v7.14.0 a request that is aborted quickly (while the `productCheck()` request is still going) properly results in the request callback (or promise resolution) including the `RequestAbortedError` instance.  However, the emitted Client "response" event does not include that `RequestAbortedError` instance.

This is a challenge for https://github.com/elastic/apm-agent-nodejs which is attempting to use the "request" and "response" Client events for APM tracing. This means APM will not report an error for a quick abort.

## To Reproduce

1. Save this as "esfoo.js"

```js
const { Client } = require('@elastic/elasticsearch')
var client = new Client({
  node: 'http://localhost:9200',
  auth: { username: 'admin', password: 'changeme' }
})

client.on('response', function (err, event) {
  console.log('"response" event: err=', err && err.name)
})

var req = client.search(
  { body: { size: 1 } },
  function (err, _result) {
    console.log('callback: err:', err && err.name)
  }
)
setImmediate(function () {
  req.abort()
})
```

with earlier versions:

```
% npm ls @elastic/elasticsearch
elastic-apm-node@3.19.0 /Users/trentm/el/apm-agent-nodejs8
└── @elastic/elasticsearch@7.13.0

% node esfoo.js
"response" event: err= RequestAbortedError
callback: err: RequestAbortedError
```

with v7.14.0:

```
% npm ls @elastic/elasticsearch
elastic-apm-node@3.19.0 /Users/trentm/el/apm-agent-nodejs8
└── @elastic/elasticsearch@7.14.0

% node esfoo.js
"response" event: err= null
callback: err: RequestAbortedError
```

## Expected behavior

I realize the "response" event being sent here is for the `Transport.prototype.request` for the *`productCheck()`* so it might not be complete straightforward. Could the `req.abort()` be passed through to the productCheck request if it is ongoing?

## Your Environment

- *node version*: v12.22.1
- *os*: Mac

delvedor:
Hello! Bug confirmed I'll work on a fix. Thank you for reporting!

PR #1534, opened 2021-08-23
Always emit request aborted event

delvedor:
If the client is busy running an async operation, the `.abort()` call might be executed before sending the actual request. In such case, the error was swallowed, now it will always be emitted, either in the `request` or `response` event.

Closes: #1517

### DIFF
--- a/lib/Transport.js
+++ b/lib/Transport.js
@@ -182,6 +182,7 @@ class Transport {
 
     const makeRequest = () => {
       if (meta.aborted === true) {
+        this.emit('request', new RequestAbortedError(), result)
         return process.nextTick(callback, new RequestAbortedError(), result)
       }
       meta.connection = this.getConnection({ requestId: meta.request.id })
--- a/test/acceptance/product-check.test.js
+++ b/test/acceptance/product-check.test.js
@@ -20,7 +20,7 @@
 'use strict'
 
 const { test } = require('tap')
-const { Client } = require('../../')
+const { Client, errors } = require('../../')
 const {
   connection: {
     MockConnectionTimeout,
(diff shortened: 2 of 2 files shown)


trentm:
This works for me with the node.js APM agent, where "works" here is that the APM instrumentation of `@elastic/elasticsearch` will notice and capture an APM "error" event for this `RequestAbortedError`.

The mechanism is slightly different than described in #1517, because the `RequestAbortedError` is now being captured in the "request" event rather than the "response" event... which makes sense because the `POST /_search` request isn't even started when the `req.abort()` happens during the productCheck.

### test notes

I installed this feature branch:

```
npm install github:elastic/elasticsearch-js#fix-1517
```

My apm-using test script that will start a transaction and make a `client.search()` call:

```js
const apm = require('./').start({
  captureExceptions: false,
  logUncaughtExceptions: true,
  captureSpanStackTraces: false,
  stackTraceLimit: 3,
  apiRequestTime: 3,
  metricsInterval: 0,
  cloudProvider: 'none',
  centralConfig: false,
  // ^^ Boilerplate config above this line is to focus on just tracing.
  serviceName: 'esfoo'
})

var client = new Client({
  auth: { username: 'admin', password: 'changeme' }
})

client.on('response', function (err, event) {
  console.log('"response" event: err=', err && err.name)
})

apm.startTransaction('t0')
var req = client.search(
  { body: { size: 1 } },
  function (err, _result) {
    console.log('callback: err:', err && err.name)
    apm.endTransaction()
  }
)
setImmediate(function () {
  req.abort()
})
```

Running that:

```
% node esfoo.js
"response" event: err= null
callback: err: RequestAbortedError
{"log.level":"info","@timestamp":"2021-08-23T18:45:17.204Z","log":{"logger":"elastic-apm-node"},"ecs":{"version":"1.6.0"},"message":"Sending error to Elastic APM: {\"id\":\"759c2d784e4e716e8b98e00f080dd86f\"}"}
```

That captured error:

```json
{
    "error": {
        "id": "759c2d784e4e716e8b98e00f080dd86f",
        "timestamp": 1629744317196000,
        "parent_id": "3fab07fbe0a7fe76",
        "trace_id": "e6e2fecf298ed8993795200f696c9556",
        "transaction_id": "71c41b9ecdbaca66",
        "transaction": {
            "type": null,
            "sampled": true
        },
        "context": {
            "user": {},
            "tags": {},
            "custom": {}
        },
        "exception": {
            "message": "Request aborted",
            "type": "RequestAbortedError",
            "handled": true,
            "attributes": {
                "name": "RequestAbortedError",
                "message": "Request aborted"
            },
            "module": "@elastic/elasticsearch",
            "stacktrace": [
                ...
            ]
        },
        "culprit": "makeRequest (node_modules/@elastic/elasticsearch/lib/Transport.js)"
    }
}
```

```

trentm (approved):
Works with the node.js APM agent. Thanks!

PR #1540, opened 2021-08-24
[DOCS] Fix double backticks for inline code snippets

ba0708:
Fixed the inline code snippets looking funky on the following page:
https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/advanced-config.html#_migrate_to_v8

Caused by double backticks being used instead of single backticks.


delvedor (approved):
Thanks!

PR #1555, opened 2021-09-21
Show socket local/remote address in case of ECONNRESET

delvedor:
As titled.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -115,7 +115,10 @@ class Connection {
       this._openRequests--
       let message = err.message
       if (err.code === 'ECONNRESET') {
-        message += ` - Local: ${request.socket.localAddress}:${request.socket.localPort}, Remote: ${request.socket.remoteAddress}:${request.socket.remotePort}`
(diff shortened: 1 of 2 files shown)


mshustov (approved):

lib/Connection.js
```diff
@@ -113,7 +113,11 @@ class Connection {
     const onError = err => {
       cleanListeners()
       this._openRequests--
-      callback(new ConnectionError(err.message), null)
+      let message = err.message
+      if (err.code === 'ECONNRESET') {
+        message += ` - Local: ${request.socket.localAddress}:${request.socket.localPort}, Remote: ${request.socket.remoteAddress}:${request.socket.remotePort}`
```
mshustov, line 118: TS definitions day `remoteAddress` and `remotePort` are optional. Should we fall back to a default value? `"unknown"`, for example

delvedor: Added a fallback, but as far as I know it should never happen, in case of ECONNRESET those properties should always be there.

PR #1596, opened 2021-11-03
Aggregations should be a generic in responses

delvedor:
As titled, this allows to write code like this:
```ts
import * as T from '@elastic/elasticsearch/lib/api/types'

const client = new Client({
  node: 'http://localhost:9200'
})

interface Document {
  world: number
}

interface Aggregation {
  hello: T.AggregationsValueAggregate
}

async function run () {
  const result = await client.search<Document, Aggregation>({
    index: 'foo',
    query: { match_all: {} },
    aggregations: {
      hello: {
        avg: {
          field: 'world'
        }
      }
    }
  })

  result.aggregations.hello // autocompletes to AggregationsValueAggregate
}
```

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -58,7 +58,6 @@ export type BulkOperationType = 'index' | 'create' | 'update' | 'delete'
 
 export interface BulkRequest<TSource = unknown> extends RequestBase {
   index?: IndexName
-  type?: Type
   pipeline?: string
   refresh?: Refresh
   routing?: Routing
@@ -87,7 +86,6 @@ export interface BulkResponseItem {
   result?: string
   _seq_no?: SequenceNumber
   _shards?: ShardStatistics
-  _type?: string
   _version?: VersionNumber
   forced_refresh?: boolean
   get?: InlineGet<Record<string, any>>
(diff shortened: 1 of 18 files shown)


src/api/types.ts
```diff
@@ -13094,13 +13094,13 @@ export interface RollupRollupSearchRequest extends RequestBase {
   size?: integer
 }
 
-export interface RollupRollupSearchResponse<TDocument = unknown> {
+export interface RollupRollupSearchResponse<TDocument = unknown, TAggregations = Record<AggregateName, AggregationsAggregate>> {
```
mshustov, line 13097: Shouldn't we enforce `TAggregations`  abide by  `Record<AggregateName, AggregationsAggregate>` interface? 
`TAggregations extends Record<AggregateName, AggregationsAggregate> = Record<AggregateName, AggregationsAggregate>`?
However,`AggregationsAggregate` is a union of types, I'm not sure it won't disrupt API consumers' code. We can do it in a follow-up.

mshustov (approved):
The overall approach looks good to me. Curious to hear @dgieselaar since the APM team has to work around the lack of `aggregations`  typed parameter in the `SearchResponse` interface

dgieselaar:
Works for me! Small change on our side.

PR #1616, opened 2022-02-06
Fix onFlushTimeout timer not being cleared when upstream errors

rprovodenko:
Hey
Found this bug when debugging a memory leak. Always close resources in `finally`. In this particular case I was getting a lot of errors in an upstream stream and this timer wasn't being released, meaning that GC couldn't clean the es client instance, causing a memory leak.


delvedor (commented):
Hello, great catch! 🚀 
Can you sign the CLA?

rprovodenko:
It's signed :)

rprovodenko:
Anything else I need to do?

PR #1643, opened 2022-02-28
Replace deprecated ssl to tls

adriaanvanrossum:
It took me a few hours to realize the docs were not up to date.

> ### Rename `ssl` option to `tls`
> 
> *Breaking: Yes* | *Migration effort: Small*
> 
> People usually refers to this as `tls`, furthermore, internally we use the tls API and Node.js refers to it as tls everywhere.
> [source,js]
> 
> ```js
> // before
> const client = new Client({
>   node: 'https://localhost:9200',
>   ssl: {
>     rejectUnauthorized: false
>   }
> })
> 
> // after
> const client = new Client({
>   node: 'https://localhost:9200',
>   tls: {
>     rejectUnauthorized: false
>   }
> })
> ```

https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/8.0/changelog-client.html#_rename_ssl_option_to_tls

### DIFF
--- a/docs/connecting.asciidoc
+++ b/docs/connecting.asciidoc
@@ -157,11 +157,9 @@ const client = new Client({
 
 Without any additional configuration you can specify `https://` node urls, and 
 the certificates used to sign these requests will be verified. To turn off 
-certificate verification, you must specify an `ssl` object in the top level 
-config and set `rejectUnauthorized: false`. The default `ssl` values are the 
-same that Node.js's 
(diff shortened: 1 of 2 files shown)


adriaanvanrossum:
Sorry, was looking at the current documentation (https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/client-connecting.html) which links to version 7.16 while 8.0.0 is released.

PR #1672, opened 2022-04-03
Fix non-working example

pragone:
the call as depicted with `index` and `operations` fails. The second parameter should be `body` instead. I had to Google examples for this.


pragone:
I've now signed the CLA, not sure if the pipeline will pick it up

PR #1732, opened 2022-07-13
Allow document to be overwritten in `onDocument` iteratee of bulk helper

### LINKED ISSUE #1209: Bulk helper document manipulation
Currently, you can't manipulate the document that gets passed to the bulk helper, if you need to do that, you should pipe the datasource in a transform stream or inside an async generator before giving it to the bulk helper:
```js
const { createReadStream } = require('fs')
const split = require('split2')

async function * manipulate (stream) {
  for await (const chunk of stream) {
    // chunk manipulation
    yield chunk
  }
}

const stream = createReadStream('./dataset.ndjson').pipe(split())

const result = await client.helpers.bulk({
  datasource: manipulate(stream),
  onDocument (doc) {
    return {
      index: { _index: 'my-index' }
    }
  }
})
```

While the solution above works perfectly, it does not support the case where the document contains metadata that does not need to be indexed, such as the index name for example:
```js
const { createReadStream } = require('fs')
const split = require('split2')

// assuming the document has the following shape:
// { index: string, id: string, body: object }

const result = await client.helpers.bulk({
  datasource: createReadStream('./dataset.ndjson').pipe(split(JSON.parse)),
  onDocument (doc) {
    return {
      index: { _index: doc.index, _id: doc.id }
    }
  }
})
```
In the code above we are indexing the document metadata as well.
For solving this problem, we could introduce another callback, to allow the document manipulation after the creation of the bulk action:
```js
const { createReadStream } = require('fs')
const split = require('split2')

// assuming the document has the following shape:
// { index: string, id: string, body: object }

const result = await client.helpers.bulk({
  datasource: createReadStream('./dataset.ndjson').pipe(split(JSON.parse)),
  onDocument (doc) {
    return {
      index: { _index: doc.index, _id: doc.id }
    }
  },
  transformDocument (doc) {
    return doc.body
  }
})
```
Another solution, but it could cause a breaking change in some cases (which is not a big deal given that the helpers are still experimental):
```js
const { createReadStream } = require('fs')
const split = require('split2')

// assuming the document has the following shape:
// { index: string, id: string, body: object }

const result = await client.helpers.bulk({
  datasource: createReadStream('./dataset.ndjson').pipe(split(JSON.parse)),
  onDocument (doc) {
    return {
      index: { _index: doc.index, _id: doc.id },
      document: doc.body
    }
  }
})
```

robdasilva:
Allows the document to be submitted to be altered in the `onDocument` iteratee of the `bulk` method of the Elasticsearch Client Helpers.

Fixes #1209


robdasilva:
@delvedor Is this PR something we could merge? Anything I can do to speed up/help with the process?

JoshMock:
Thanks for this fix @robdasilva. @delvedor is no longer the maintainer of this library, and I am taking over. I'll try to take a look soon to make sure this looks good.

robdasilva:
Thanks for the heads-up, @JoshMock. Let me know, if anything.

JoshMock:
jenkins test this

src/helpers.ts
```diff
@@ -90,7 +90,9 @@ interface DeleteAction {
   delete: T.BulkDeleteOperation
 }
 
-type UpdateAction = [UpdateActionOperation, Record<string, any>]
+type CreateAction = CreateActionOperation | [CreateActionOperation, unknown]
+type IndexAction = IndexActionOperation | [IndexActionOperation, unknown]
+type UpdateAction = [UpdateActionOperation, T.BulkUpdateAction]
```
JoshMock, line 95: Nice catch, not sure why this was `Record` before. :thinking:

JoshMock (approved):
I like this approach, even more than the two proposals in https://github.com/elastic/elasticsearch-js/issues/1209, because it is a non-breaking change to the `onDocument` return type! Everyone wins.

I did some local tests with it and it seems to work as advertised, so I'll merge shortly, add some docs to reflect the improvement, and try to get it out in 8.8.2 and 8.9.0.

PR #1750, opened 2022-08-12
Update APIs to 8.4-SNAPSHOT

sethmlarson:


rudolf (approved):
Not aware of any Kibana consumers using `.Internal`

Issue #1751, opened 2022-08-14
Issue with the bulk helper handling errors when a delete action is combined with other actions


karlriis:
## 🐛 Bug Report
When a bulk helper call contains a delete action combined with some other actions (e.g. update), and when one of the actions errors, then it is possible that the `onDrop()` callback returns incorrect information about the erroring operation/document, or a deserialization error occurs.

### Detailed explanation with background
The `tryBulk()` function in the bulk helper makes a bulk API call with multiple operations and documents (contained in `bulkBody`), and then loops over the response items. When a response item contains an error, then that item is matched with the corresponding request item from `bulkBody` to pass the problematic operation/document to the `onDrop` callback. 
https://github.com/elastic/elasticsearch-js/blob/4ebffbc0e895cec77beb3c585ea46b41391edd4c/src/helpers.ts#L811
  
To map the erroring response item to the request item, a variable named `indexSlice` is derived based on the response items loop counter. The logic for the calculation of the variable seems incorrect in the case where there are delete operations mixed with others. 
https://github.com/elastic/elasticsearch-js/blob/4ebffbc0e895cec77beb3c585ea46b41391edd4c/src/helpers.ts#L833

For example, if we create a bulk request with a delete and an update operation, then the request body will be something like this (1 item for delete, 2 items for update):
```
{ delete: { _index: '...', _id: '...' } }
{ update: { _index: '...', _id: '...' } }
{ doc: { ... } }
```
The response to this will contain only two items, one for each operation. Let's say that the update failed, in which case the second item in the response is going to contain an error. To map it to the request item, the `indexSlice` is derived as `indexSlice = 1*2`. Now, when gathering the operation and document from `bulkBody` to return error information through `onDrop()`, it is done as such:
https://github.com/elastic/elasticsearch-js/blob/4ebffbc0e895cec77beb3c585ea46b41391edd4c/src/helpers.ts#L849-L853

Which for the current example will be:
```
operation: serializer.deserialize(bulkBody[2]),
document: serializer.deserialize(bulkBody[3])
```

This is incorrect as `bulkBody[2]` contains the document instead of the operation, and `bulkBody[3]` does not exist at all. The latter causes a deserialization error.

If there would have been an additional successful action after the update, then the deserialization wouldn't have occurred, but the `indexSlice` would still be shifted by one and incorrect information would be returned for `operation` and `document` fields in `onDrop()`.

## To Reproduce

[...]

PR #1759, opened 2022-09-01
fix: keep track of indexSlice explicitly in bulk helper response handling loop

karlriis:
Closes https://github.com/elastic/elasticsearch-js/issues/1751

Fixes the aforementioned issue where a bulk helper call can return wrong information about the erroring operation/document via `onDrop()`, or cause a deserialization error when the request contains a delete action combined with any other action.

The fix is to keep track of `indexSlice` separately in the response handling loop of the `tryBulk` method, instead of deriving it based on the loop counter. The previous approach didn't account for different types of operations which were handled in previous iterations of the loop.

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -823,13 +823,13 @@ export default class Helpers {
             }
             const retry = []
             const { items } = result
+            let indexSlice = 0
             for (let i = 0, len = items.length; i < len; i++) {
               const action = items[i]
               const operation = Object.keys(action)[0]
               // @ts-expect-error
               const responseItem = action[operation as keyof T.BulkResponseItemContainer]
               assert(responseItem !== undefined, 'The responseItem is undefined, please file a bug report')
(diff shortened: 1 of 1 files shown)


karlriis:
@JoshMock Thanks, rebased against latest main now

JoshMock:
jenkins test this

JoshMock:
buildkite test this

abendi:
@JoshMock any updates on getting this over the line?

JoshMock (approved):
Sorry for the delay @karlriis! This looks right. Merging, and will add a test to verify the change. :sunglasses:

JoshMock:
jenkins test this

JoshMock:
Tests are failing on Node 14.x for unrelated reasons that I can fix elsewhere. (14.x is also EOL but we'll address that in https://github.com/elastic/elasticsearch-js/issues/1855.)

PR #1765, opened 2022-09-26
Fix docs URLs to use '8.5' instead of 'master'

sethmlarson:
Backport of https://github.com/elastic/elasticsearch-js/pull/1764


szabosteve (approved):
URL changes LGTM.

PR #1780, opened 2022-09-27
Typo: missing ','

freiit:


sethmlarson:
Thanks for this, can you use a base of `main` so this fix can land in the "to be released" version of the library?

freiit:
see here: https://github.com/elastic/elasticsearch-js/pull/1783

PR #1792, opened 2022-11-13
Add module type in package.json

sumanThapa11:
Added module type "commonjs"  in the package.json file.


sethmlarson:
This PR would need to target `main`, not a feature branch, hence the weird state of conflicts/rebasing. Going to close this one in favor of a new one if this still need to be fixed?

PR #1863, opened 2023-05-02
Buildkite pipeline for integration tests

JoshMock:
Adds a Buildkite pipeline to run client integration tests. Will drop the Jenkins job definitions once I've verified this runs as expected.


JoshMock:
Buildkite keeps failing these new jobs, telling me that the `tsc` executable is not found on `PATH`. When I run the same Docker image locally, it finds `tsc` with no problem. :man_shrugging: Will keep poking at it.

JoshMock:
Finally got it: the `docker run` in `run-repository.sh` was mapping volumes that overwrote the results of `npm install`. Doing a `COPY` of everything into the image instead of mapping volumes works like a charm.

PR #1913, opened 2023-06-08
[Backport 7.x] fix: keep track of indexSlice explicitly in bulk helper response handling loop (#1759)

Co-authored-by: Josh Mock <josh@joshmock.com>

JoshMock:
Backport of de17dc05 from #1759


JoshMock:
bash script backport fail!

PR #1925, opened 2023-06-28
fix: types.ts SnapshotGetRequest

potreco:
Changing the type that is causing error on NestJS build

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -16739,10 +16739,10 @@ export interface SnapshotGetRequest extends RequestBase {
   index_names?: boolean
   include_repository?: boolean
   sort?: SnapshotSnapshotSort
-  size?: integer
+  size?: number
   order?: SortOrder
   after?: string
-  offset?: integer
+  offset?: number
   from_sort_value?: string
   slm_policy_filter?: Name
 }


aitherios:
Just as a suggestion. This can also be sorted out by importing the integer type from the types module. Patch bellow:
```
diff --git a/api/requestParams.d.ts b/api/requestParams.d.ts
index ccd2889e830d15eed9006ddf633bc0bb5081140d..bcaaa8bbb32c735a4dc7f2915c3b2aa2472d5f11 100644
--- a/api/requestParams.d.ts
+++ b/api/requestParams.d.ts
@@ -18,6 +18,7 @@
  */
 
 import { RequestBody, RequestNDBody } from '../lib/Transport'
+import { integer } from './types'
 
 export interface Generic {
   method?: string;
```

weiwong:
Following @aitherios's suggestion, I think we should update `generateRequestTypes.js` instead, since `requestParams.d.ts` is generated using that script:

```diff
diff --git a/scripts/utils/generateRequestTypes.js b/scripts/utils/generateRequestTypes.js
index 854e9ebc..2fbf9fc6 100644
--- a/scripts/utils/generateRequestTypes.js
+++ b/scripts/utils/generateRequestTypes.js
@@ -45,6 +45,7 @@ function generate (version, api) {
  */
 
 import { RequestBody, RequestNDBody } from '../lib/Transport'
+import { integer } from './types'
 
 export interface Generic {
   method?: string;
```

JoshMock:
Thanks for the contribution @potreco. These lines of code are generated from the Elasticsearch spec, so I was able to do a more permanent fix to the generator _and_ the code in https://github.com/elastic/elasticsearch-js/pull/1927.

Issue #1924, opened 2023-06-28
typing error on requestParam.d.ts


potreco:
## 🐛 Bug Report

A type error on `SnapshotGet`, on file `requestParam.d.ts` file.
I'm opened a PR to fix this: https://github.com/elastic/elasticsearch-js/pull/1925

![MicrosoftTeams-image](https://github.com/elastic/elasticsearch-js/assets/8931135/0d5248c9-c5dc-4991-b8f6-fb1b4cb1bdb1)

Lines: 2632  and 2636, change both to `number`.

## To Reproduce

* Install this lib on a NestJS project
* Build the project

## Expected behavior

Build occur normally

## Your Environment

- *node version*: 16
- `@elastic/elasticsearch` *version*: 7.17.11
- *os*: Mac, Windows, Linux

adambiggs:
Is type checking part of the CI pipeline?

MichaelRoskin:
Im getting the same...
kids, that's what happens when you run the CI without a package-lock.json - lesson learned

JoshMock:
Thanks for the report, all. The issue likely will need to be resolved in the underlying code generator or the Elasticsearch specification that the generator uses. Once it's applied it will be published in the next 7.x release.

In the meantime, you can pin your dependencies to 7. 

There is a type-check done before publishing, but it appears to have failed silently or been bypassed somehow. I'll investigate that separately to ensure failures successfully block publishes going forward.

JoshMock:
The fix in https://github.com/elastic/elasticsearch-js/pull/1927 will resolve this issue. Unfortunately the npm package did not qualify for unpublishing, so we will be publishing version `7.17.11-patch.1` shortly.

Once that publishes, you can either pin to version `7.17.11-patch.1` or `7.17.0` until we release `7.17.12`. Versioning of the client is tied to the Elasticsearch release schedule, and we currently do not have a date for when that might release.

Thanks again for your patience, everyone. :pray:

JoshMock:
`7.17.11-patch.1` has been published to npm. Release notes around when to use the patch version are clarified on [the release page](https://github.com/elastic/elasticsearch-js/releases/tag/v7.17.11).

distracteddev:
@JoshMock Just wanted to share my experience that this is not a real solution for people because many dependencies are included by packages we do not have control over, so there is no good way to pin the package to that version across all dependents.

JoshMock:
[...]

PR #1927, opened 2023-06-29
Type declaration fix for 7.17.11

JoshMock:
Fixes https://github.com/elastic/elasticsearch-js/issues/1924 and bumps npm package version to `7.17.11-patch.1`.


JoshMock:
Also notable: skipping some unit tests that were failing will ensure that TypeScript type-checking is run during `npm test`, until we can fix those failing tests. Because they were failing and known to be failing, it was easy to ignore a failed test run that hadn't yet run a type check.

PR #1930, opened 2023-07-03
fix: add types to exports (v7.x)

Marsup:
Hi,

Enabling ESM mode with TypeScript seems to cause this error to appear: `Cannot find module '@elastic/elasticsearch' or its corresponding type declarations.(2307)`

We're not ready yet to upgrade to v8.x which is why I'm patching v7.x, but it seems this error is also present in this version, and I'm not sure how to fix it yet, maybe it's as simple as declaring `"type": "commonjs"` in the package.json?

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
   "exports": {
     ".": {
       "require": "./index.js",
-      "import": "./index.mjs"
+      "import": "./index.mjs",
+      "types": "./index.d.ts"
     },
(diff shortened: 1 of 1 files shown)


JoshMock (approved):
Good catch. Thanks @Marsup!

JoshMock:
Verified that all the failing/missing CI tests are unrelated to this change. Merging (finally)!

Issue #1937, opened 2023-07-07
MultiSearch response should include generic aggregations type


hoiekim:
## 🐛 Bug Report

Hi maintainers,

Regarding this pull request #1596 (Aggregations should be a generic in responses), I believe there was missing changes in this line
https://github.com/elastic/elasticsearch-js/blob/2375059aa36281a5b7b004e3d0838a96ddf04238/src/api/types.ts#L782
Note that `TAggregations` is not passed in `MsearchMultiSearchResult`. This results in typescript not being able to recognize aggregations type in the response of `Client.msearch<TDocument, TAggregations>`.

## To Reproduce

```js
import { Client } from "@elastic/elasticsearch";
import { AggregationsTermsAggregateBase } from "@elastic/elasticsearch/lib/api/types";

const client = new Client({ node: "http://localhost:9200" });

const reproduce = async () => {
  interface Doc {
    foo: string;
  }

  interface Aggregations {
    unique: AggregationsTermsAggregateBase<{ key: string }>;
  }

  const response = await client.msearch<Doc, Aggregations>({
    index: "foo",
    searches: [
      {},
      {
        query: { match_all: {} },
        aggregations: { unique: { terms: { field: "foo" } } }
      }
    ]
  });

  response.responses.map((e) => {
    if ("error" in r) return
    return e.aggregations?.unique.buckets;
    //                            ^^^^^^^
    // Property 'buckets' does not exist on type 'AggregationsAggregate'.
    // Property 'buckets' does not exist on type 'AggregationsCardinalityAggregate'.
  });
};
```

## Expected behavior

Should not throw compile errors

## Your Environment

- *node version*: 18.15.0
- `@elastic/elasticsearch` *version*: 8.8.1
- `typescript` *version*: 4.5.5
- *os*: MacOS

JoshMock:
Hi @hoiekim, sorry for the long delay.

I spent the day looking at this and did find one type, `MsearchMultiSearchResult`, where it can be improved to include a generic aggregation type.

However, in the case of `MsearchMultiSearchItem`, it already inherits a `ResponseBody` that includesa generic aggregation type, so there's nothing I can do there. The reason you get a TypeScript message is because `response.responses` is a `MsearchResponseItem`, which itself is a union:

https://github.com/elastic/elasticsearch-js/blob/99bcff02d5edf0e9200f3a36baf6df9f32aa6bf8/src/api/types.ts#L785

[...]

PR #1938, opened 2023-07-07
Fix: Implements generic aggregations for multisearch

hoiekim:
This PR fixes the following bug (#1937)
```js
import { Client } from "@elastic/elasticsearch";
import { AggregationsTermsAggregateBase } from "@elastic/elasticsearch/lib/api/types";

const client = new Client({ node: "http://localhost:9200" });

const reproduce = async () => {
  interface Doc {
    foo: string;
  }

  interface Aggregations {
    unique: AggregationsTermsAggregateBase<{ key: string }>;
  }

  const response = await client.msearch<Doc, Aggregations>({
    index: "foo",
    searches: [
      {},
      {
        query: { match_all: {} },
        aggregations: { unique: { terms: { field: "foo" } } }
      }
    ]
  });

  response.responses.map((e) => {
    if ("error" in r) return
    return e.aggregations?.unique.buckets;
    //                            ^^^^^^^
    // Property 'buckets' does not exist on type 'AggregationsAggregate'.
    // Property 'buckets' does not exist on type 'AggregationsCardinalityAggregate'.
  });
};
```
The error will go way with this commit

Similar PR: #1596

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -702,13 +702,13 @@ export interface MgetResponse<TDocument = unknown> {
 
 export type MgetResponseItem<TDocument = unknown> = GetGetResult<TDocument> | MgetMultiGetError
 
-export interface MsearchMultiSearchItem<TDocument = unknown> extends SearchResponseBody<TDocument> {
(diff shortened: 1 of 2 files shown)


JoshMock:
And there's the PR I saw coming. :slightly_smiling_face: I'll take a look soon.

JoshMock:
Closing. See: https://github.com/elastic/elasticsearch-js/issues/1937#issuecomment-1793024394

PR #1939, opened 2023-07-13
7.17.12 changelog

JoshMock:
Adds 7.17.12 changelog (see cbc0166c40eeb30b2ef3a18a304c6d407267fe3f), as well as backporting lots of commits that did not get applied to 7.x.


JoshMock:
@picandocodigo @ezimuel I opened this as a draft because I wanted to verify what should go in the 7.x branch. My assumption is that it's the same purpose as `main` is currently for 8.x. If that's the case, a bunch of commits for 7.16 and 7.17 were never merged into 7.x, which is why I'm adding them here.

picandocodigo:
@JoshMock I'll check in more detail tomorrow, but back when we migrated to `8.x`, the idea was to either rename `7.x` to `7.17`, or delete `7.x` and only keep maintaining `7.17` from the `7.x` series. Maybe that's why `7.x` was behind `7.17`, it just hadn't been deleted?

JoshMock:
That makes sense! It won't hurt to operate under that assumption for another day so I can get publish 7.17.12 tomorrow. :slightly_smiling_face: We can always backport all these commits to 7.x later if it's useful.

PR #1945, opened 2023-07-16
Update README.md

ibrahimkazimov:
Changed search function in readme because it requires body parameter. Otherwise returns error "Missing required parameter body"

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,9 +128,12 @@ async function run () {
   // Let's search!
   const result= await client.search({
     index: 'game-of-thrones',
-    query: {
-      match: { quote: 'winter' }
+    body: {
+      query: {
+        match: { quote: 'winter' }
+      }
     }
+   
   })
 
   console.log(result.hits.hits)


ibrahimkazimov:
Here is the error I tried to solve:
Screenshot 2023-07-16 at 4 01 42 PM


It actually says contains unrecognized parameter query.

JoshMock:
In 8.x, `query` is a valid argument in the `client.search` API function, as you can see [here](https://github.com/elastic/elasticsearch-js/blob/2375059aa36281a5b7b004e3d0838a96ddf04238/src/api/api/search.ts#L49).

I also verified with a simple example that using `query` does not return a `ResponseError` when using Elasticsearch 8.8.0 with version 8.8.1 of the JavaScript client on Node.js v20.3:

```javascript

const client = new Client({
  auth: {
    username: 'elastic',
    password: 'changeme'
  }
})

async function doSearch() {
  const result = await client.search({
    index: 'foobar',
    query: { match: { foo: 'bar' }}
  })
  console.log(result)
}

doSearch()
  .then(() => console.log('done'))
  .catch(err => console.error(err))
```

@ibrahimkazimov Can you share what version of Elasticsearch, the client, and Node.js you're using? Please note that a [major-version difference between your client and server versions](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/installation.html#js-compatibility-matrix) can lead to unexpected behavior. That may not be the issue here, but it could be a contributing factor.

ibrahimkazimov:
I created an account on [elastic.co](https://www.elastic.co/). And then I tried to connect to the cloud. I npm installed @elastic/elasticsearch version 7.17.12. 
This is the version that gets downloaded when you type npm install @elastic/elasticsearch

JoshMock:
Sorry about that. Publishing 7.17.12 last week unintentionally set the `latest` tag. If you uninstall it, `npm install @elastic/elasticsearch` will now correctly install 8.8.1 instead.

PR #1954, opened 2023-07-20
Update user agent format

JoshMock:
I fixed the user-agent to not be `elastic-transport-js` in https://github.com/elastic/elasticsearch-js/pull/1865, but the format was slightly inconsistent with other clients and the transport user-agent itself. Nothing drastic that will break client usage telemetry, just keeping things consistent.


picandocodigo (approved):
LGTM! Checking out how my client did it after looking at this code, I think I found a potential issue with my client's user agent code :raised_hands:

Issue #1828, opened 2023-03-10
`flags` and `group` in `AnalysisPatternTokenizer` should be marked as optional.


LvChengbin:
It's not uncommon that somebody already opened an issue or in the best case it's already fixed but not merged. That's the reason why you should [search](https://github.com/elastic/elasticsearch-js/issues) at first before submitting a new one.

**Please read this entire template before posting any issue. If you ignore these instructions
and post an issue here that does not follow the instructions, your issue might be closed,
locked, and assigned the `not reproducible` label.**

## 🐛 Bug Report

https://github.com/elastic/elasticsearch-js/blob/948f44a7a0dce274850c768dae7dbb898dbcd988/src/api/types.ts#L4213

```ts
export interface AnalysisPatternTokenizer extends AnalysisTokenizerBase {
  type: 'pattern'
  flags: string
  group: integer
  pattern: string
}
```

Both `flags` and `group` property should be marked as `optional`.

From the doc [Analysis Pattern Tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pattern-tokenizer.html), we can see that the `flags` is not a required param and `group` has a default value.

## Your Environment

- *node version*: 6,8,10
- `@elastic/elasticsearch` *version*: >=7.0.0
- *os*: Mac, Windows, Linux
- *any other relevant information*

PR #1967, opened 2023-08-09
Fix AnalysisPatternTokenizer definition (#1828)

dolivier83:
Closes https://github.com/elastic/elasticsearch-js/issues/1828 by changing `flag`, `group` and `pattern` as optional properties of `AnalysisPatternTokenizer` type.

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -4495,9 +4495,9 @@ export interface AnalysisPatternReplaceTokenFilter extends AnalysisTokenFilterBa
 
 export interface AnalysisPatternTokenizer extends AnalysisTokenizerBase {
   type: 'pattern'
-  flags: string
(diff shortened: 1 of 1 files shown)


JoshMock (approved):
I had to check [the source](https://github.com/elastic/elasticsearch/blob/d7920a315cd61b6ff88bc137e343df321cea6ab5/modules/analysis-common/src/main/java/org/elasticsearch/analysis/common/PatternTokenizerFactory.java) verify that all three values can indeed be null. Good catch!

I've also opened https://github.com/elastic/elasticsearch-specification/pull/2232 to ensure this change gets applied to all our clients.

Thanks for the contribution!

JoshMock:
jenkins test this

PR #1971, opened 2023-08-15
Remove unnecessary code

wonseop:
Function 'buildPath' is defined without arguments, so I removed an unnecessary argument from the function call. 
And I removed a parameter 'param' of function 'genRequiredChecks'.


JoshMock (approved):
Thanks for the contribution. These scripts are mostly unused now, as we've moved code generation into a separate project. Can't hurt to merge the improvement, though. :black_heart:

Issue #1562, opened 2021-10-05
Bulk helper semaphore handling results in hanging await for long-running requests (exceeding flushInterval)


brentmjohnson:
## 🐛 Bug Report

Bulk helper hangs forever when flushInterval is exceeded while iterator is already awaiting semaphore.

## To Reproduce

1. Run below code against a test cluster - should complete successfully
2. Simulate long running server-side operation exceeding configured flushInterval.  Multiple ways to do this but one way is to modify the compiled Helpers.js with the following:

```js
client.bulk(Object.assign({}, bulkOptions, { body: bulkBody }), reqOptions, async (err, { body }) => {
await new Promise(resolve => setTimeout(resolve, flushInterval));
if (err) return callback(err, null)
```

3. Re-run below code and watch the hanging await caused by onFlushTimeout() invoked on a payload already awaiting semaphore()

Paste your code here:

```js
'use strict'
const util = require("util")
const { Readable } = require('stream');
const { Client } = require('@elastic/elasticsearch');

async function* generator() {
    let i = 0
    while (i < 10) {
        await new Promise(resolve => setTimeout(resolve, 1 * 1000));
        yield { i: i++ };
    }
};

const readableStream = Readable.from(generator());

const elasticClient = new Client({
    [TESTCLUSTER]
});

(async () => {
    const bulkHelper = elasticClient.helpers.bulk({
        flushBytes: 43,
        concurrency: 1,
        datasource: readableStream,
        onDocument(doc) {
            console.log(doc)
            return {
                index: { _index: 'semaphoretest' }
            }
        },
        onDrop(doc) {
            console.log(doc);
        }
    }).catch((err) => {
        console.error(err);
    });

    while (util.inspect(bulkHelper).includes('pending')) {
        await new Promise(resolve => setTimeout(resolve, 1 * 1000));
        console.log('...waiting');
    }

    console.log(await bulkHelper);
})();
```

## Expected behavior

Bulk helper awaits gracefully for queued requests to complete, error, or timeout.

Paste the results here:

```js
{ i: 0 }
...waiting
{ i: 1 }
...waiting
...waiting
...waiting
...[forever]
```

## Your Environment

- *node version*: v14.17.6
- `@elastic/elasticsearch` *version*: >=7.15.0
- *os*: Linux

p1x44r:
[...]

PR #2027, opened 2023-10-03
Fix hang in bulk helper semaphore when server responses are slower than flushInterval

JoshMock:
A possible fix for https://github.com/elastic/elasticsearch-js/issues/1562.

The failing state in the above issue is reached when a server's response times are slower than `flushInterval`. What happens in this situation, in this order:

1. The [`flushBytes` block](https://github.com/elastic/elasticsearch-js/blob/65580b0a2d59179ea9d63be315051588d0b25ab1/src/helpers.ts#L649-L655) awaits `semaphore()`
2. While awaiting, `onFlushTimeout` is triggered
3. [`onFlushTimeout` "steals"](https://github.com/elastic/elasticsearch-js/blob/65580b0a2d59179ea9d63be315051588d0b25ab1/src/helpers.ts#L684-L690) the bulk body rows that the `flushBytes` block is waiting to send by emptying `bulkBody` before awaiting a semaphore
4. When the `flushBytes` block's semaphore resolves, it tries to `send(bulkBody)` but `bulkBody` is empty

By having `flushBytes` set its blocks aside before awaiting the semaphore, like `onFlushTimeout` already does, it appears to solve the situation where a Promise sits unresolved indefinitely.

@delvedor I'm particularly interested in your thoughts on this solution, since you wrote the helper and chose the semaphore strategy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@elastic/elasticsearch",
-  "version": "8.9.1",
-  "versionCanary": "8.9.1-canary.1",
+  "version": "8.10.1",
+  "versionCanary": "8.10.1-canary.1",
   "description": "The official Elasticsearch client for Node.js",
   "main": "index.js",
   "types": "index.d.ts",
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -648,10 +648,11 @@ export default class Helpers {
 
         if (chunkBytes >= flushBytes) {
           stats.bytes += chunkBytes
-          const send = await semaphore()
-          send(bulkBody.slice())
+          const bulkBodyCopy = bulkBody.slice()
           bulkBody.length = 0
           chunkBytes = 0
+          const send = await semaphore()
+          send(bulkBodyCopy)
         }
       }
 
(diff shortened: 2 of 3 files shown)


JoshMock:
I've published this to npm as `@elastic/elasticsearch-canary@8.10.1-canary.1` so that it can be tested by a few affected people.

pquentin (dismissed):
Your explanation makes sense, but I'm unable to map it back to the code, sorry.

JoshMock:
@pquentin Thanks. :black_heart: Mostly tagged dynamic clients team members for visibility because this bug came up in conversation in our Monday team meeting.

It still bends my mind a bit, too, and I can't find a clear explanation for why an empty `bulkBody` causes the promise to not resolve. But that's a relatively small gap to jump, especially when the fix does solve the problem in every scenario I've been able to recreate. :laughing:

alkozayMoein (approved):

ezimuel (dismissed):
The solution proposed to store the `bulkBody` before the `await semaphore()` makes sense.  I'm wondering if the issue came because of the [slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) usage, since it's create a shallow copy of the array.

pquentin (approved):

PR #2137, opened 2024-02-23
Upgrade transport to 8.4.1

JoshMock:


trentm:
Not sure if it matters, but I noticed that the 8.13.0 release (and the "8.13" branch) does *not* have this.  It means that the v8.12.2 release has `"@elastic/transport": "^8.4.1",`, but the v8.13.0 release has `"@elastic/transport": "^8.4.0",`.

I think the "8.13" branch was created from the commit *before* this one to main (commit 1607a0d3).

JoshMock:
Good catch, @trentm. I'll backport to 8.13 and get a patch release out soon.

PR #2154, opened 2024-03-07
Update bulk.asciidoc change 'date' property to 'time' to match index creation

SuperGingie:
In the code example, the index is created with a property called 'time' but the dataset has a field called 'date.' Updating the 'date' field to match the 'time' property of the index.


JoshMock (approved):
good catch! thanks for the contribution.

JoshMock:
buildkite test this

PR #2208, opened 2024-04-07
Add `Date` to `DateMath` type

ramikg:
The following code produces a TypeScript error but it does work (if the error is ignored):
```js
elastic.search({
    index: myIndex,
    query: {
        range: { timestamp: { lte: new Date() } }
    },
});
```

The error is "_TS2769: No overload matches this call_", as `QueryDslRangeQuery` expects `Datemath | double` (i.e. `string | number`) for the `gt`/`gte`/`lt`/`lte` properties.

What do you think about changing `DateMath` to be `string | Date`?


JoshMock:
@ramikg Nice catch. JavaScript automatically serializes it to a string in the necessary places, so that does totally work! I also fixed the upstream code generator logic to persist your change, and to add it to the `DateTime` type as well. :clap:

JoshMock (approved):

ramikg:
Thank you!

PR #2209, opened 2024-04-08
`fields` added to `top_hits` aggregation

semd:
The following query works and returns the `fields` information requested in the `top_hits` aggregation. However, the typescript definition for top hits aggregations (`AggregationsTopHitsAggregation`) is missing the `fields` entry. 

```js
{
  "aggregations": {
    "theAggName": {
      "top_hits": {
        "fields": [
          {
            "field": "@timestamp",
            "format": "strict_date_optional_time"
          },
          "_index"
        ]
      }
    }
  }
}
```
So typescript throws the following error:
`Object literal may only specify known properties, but 'fields' does not exist in type 'AggregationsTopHitsAggregation'. Did you mean to write 'field'?`

This PR adds the `fields` property to the `AggregationsTopHitsAggregation` type with:

`fields?: (QueryDslFieldAndFormat | Field)[]`

 I've tested it works using the `QueryDslFieldAndFormat` and `string` types, and the array is required.

### DIFF
--- a/src/api/types.ts
+++ b/src/api/types.ts
@@ -4062,6 +4062,7 @@ export interface AggregationsTopHitsAggregate extends AggregationsAggregateBase 
 export interface AggregationsTopHitsAggregation extends AggregationsMetricAggregationBase {
   docvalue_fields?: Fields
   explain?: boolean
+  fields?: (QueryDslFieldAndFormat | Field)[]
   from?: integer
   highlight?: SearchHighlight
   script_fields?: Record<string, ScriptField>


semd:
fwiw, this comes from a Kibana PR comment https://github.com/elastic/kibana/pull/178893#discussion_r1544948431

JoshMock:
@semd Thanks for this. It's best we look at [the underlying specification](https://github.com/elastic/elasticsearch-specification) to fix missing types like this, so that all clients benefit from the fix.

I'll see about fixing the spec soon, unless you happen to get to it before I can.

semd:
@JoshMock That sounds good. I didn't know about this process, I just saw the type missing and jumped into the fix. 
We can follow the conventional way, I already created a local type to solve the problem in Kibana. 
Please let me know when the spec is fixed so I can update our side. Thanks!

JoshMock:
I've opened https://github.com/elastic/elasticsearch-specification/pull/2522 to solve this in the spec. Code generation should pick it up and include it in the next release.

PR #2284, opened 2024-06-08
trying to get generate-doc-examples.js working

anniegale9538:
Hi Hello!

I've been contributing to the Python doc examples in the Elasticsearch-py repo (with Quentin Pradet's help) and am trying to get the doc examples for the other languages available!

Disclaimer: I'm not a JS or TypeScript developer, but wanted to take a whack at getting a code change up.

The alternative_reports.json file that used to live in the client_teams repo looks to have moved to the client-fliight-recorder repo and is now using the parsed_alternative_report.json file to generate the doc examples.

I was able to get this working making a few changes to the generate_doc_examples.js file.

I was unable to get it working used node, but was able to get it running using 'npx tsx' (You may hate this and I'm sorry!)

I was able to get the doc examples generating locally, but didn't want to add them and clutter this initial PR.


JoshMock (approved):
LGTM! Thanks for making the change. :clap:

package.json
```diff
@@ -73,6 +74,7 @@
     "rimraf": "^3.0.2",
     "semver": "^7.3.7",
     "split2": "^4.1.0",
+    "standard": "^17.1.0",
```
JoshMock, line 77: Actually, looks like adding `standard` is making the linter fail. We already have `ts-standard` installed, so it would be better to import and use that instead.

JoshMock (changes requested):

package.json
anniegale9538: Okay, I THINK I got it!

scripts/generate-docs-examples.js
```diff
@@ -142,7 +140,15 @@ console.log(response${getResponsePostfix(i)})
...
-  const { results } = standard.lintTextSync(code, { fix: true })
+  const { results } = await new Promise((resolve, reject) => {
+    tsStandard.lintText(code, { fix: true }, (err, result) => {
+      if (err) {
+          reject(err)
+      } else {
+          resolve(result)
+      }
+    })
+  })
```
JoshMock, line 151: I don't think we need to wrap this explicitly in a Promise. Since you made `generateAsciidoc` an async function, any errors thrown by `lintText` will be thrown just like any other exception would be.

```suggestion
  const { results } = await tsStandard.lintText(code, { fix: true })
```

JoshMock (changes requested):

JoshMock:
Going to close this, as we have a new strategy for generating examples that's about to land. I already have it running locally, but some of the code isn't ready to publish yet.

PR #2338, opened 2024-08-14
Update request-converter version in package.json

miguelgrinberg:
There is no 8.16 version of the request converter yet, changing it to 8.15.


JoshMock (approved):
Good catch. I fixed this in main, but incidentally in a PR that I didn't backport to 8.15!

PR #2432, opened 2024-11-04
Add `_id` to the result of `helpers.search`

ramikg:
# Overview
This PR adds the `_id` field to the object returned by `helpers.search`.

# Motivation
A vast majority of my use cases require knowing the IDs of the documents being fetched, which is why I end up using instead the regular `search` function instead of the `helpers.search` handy helper.

# Additional considerations

At first I thought about adding an extra parameter to the function (`extraOptions`?), but after some consideration, I'm not sure that changing the existing behavior would be disruptive.  
Still, I don't mind re-implementing this feature in any other manner.

## Some notes
1. `_id` is an illegal field name, meaning `_source` will never include it.
2. Here is the `@es_quirk` I'm referring to in the comments (exists since version 8.14.0):
    ```
   /**
      * @es_quirk '_id' is not available when using 'stored_fields: _none_'
      * on a search request. Otherwise the field is always present on hits.
   */
   ```
    (In the code I rather not quote it.)

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -28,6 +28,7 @@ import { errors, TransportResult, TransportRequestOptions, TransportRequestOptio
 import { Table, TypeMap, tableFromIPC } from '@apache-arrow/esnext-cjs'
 import Client from './client'
 import * as T from './api/types'
+import { Id } from './api/types'
 
 export interface HelpersOptions {
   client: Client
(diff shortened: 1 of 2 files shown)


JoshMock (approved):
This makes complete sense. Thanks for the contribution!

ramikg:
Thank you!

Issue #2466, opened 2024-11-14
ESM import issue (replicated with Deno v2 and Node 23)


eturino:
### 🐛 Bug report

With v8.16.0 in a deno v2 project, we run into the following problem when running:

```
error: Could not resolve 'npm:@elastic/elasticsearch@8.16.0'.

Caused by:
    [ERR_INVALID_PACKAGE_TARGET] Invalid "exports" main target {"require":"./index.js"} defined in the package config /workspaces/entities-api/node_modules/.deno/@elastic+elasticsearch@8.16.0/node_modules/@elastic/elasticsearch/package.json imported from 'file:///workspaces/entities-api/src/search-api/es-client.ts'; target must start with "./"
```

For what I can see, the issue is in the `exports` section of the package.json. Probably should be `"exports": { ".": "./index.js" },` (replace "require" with ".")

The issue is not present with 8.15.x versions, which are commonjs packages entirely.

### To reproduce

In a deno v2 (potentially also other ESM runtimes), install version 8.16.0 and run your app and/or any tests that uses a file that imports the es library.

### Expected behavior

Should be imported and run as expected

### Node.js version

Deno v2

### @elastic/elasticsearch version

8.16.0

### Operating system

Debian bookworm

ffMathy:
This is not just Deno. Just happened on Node 23 as well for me. The new *.16 release is broken.

eturino:
I expected so, but I only tried it in deno.

ffMathy:
@JoshMock I am mentioning you here, as I believe you'd be interested in knowing that the latest version doesn't work at all. It can't even be imported.

ajkraus04:
This is crazy, how not a single javascript user can use elastic right now, this should be a urgent issue!!!!!!!

sankoshine:
So it's really a bug? Any workaround on this?

ajkraus04:
Just down grade to 7.16.10 or whatever the prior major version was and that will fix issue for now.

zurmokeeper:
@eturino You can refer to this PR #2467  for a temporary solution, or downgrade to a version before 8.16.0.

hitmands:
```
[commonjs--resolver] Failed to resolve entry for package "@elastic/elasticsearch". The package may have incorrect main/module/exports specified in its package.json: No known conditions for "." specifier in "@elastic/elasticsearch" package
transforming (51) ../index.ts
```

A quick turnaround would be greatly appreciated.

[...]

PR #2467, opened 2024-11-15
Add Esm export support

zurmokeeper:
Add Esm export support  #2466


JoshMock:
closing due to no CLA signature, and needing to patch and release this sooner rather than later.

PR #2573, opened 2025-01-17
Drop default 30-second timeout

JoshMock:
Drops the default 30-second timeout from all HTTP requests, to better align with [Elasticsearch best practices for HTTP clients](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html#_http_client_configuration).

Timeouts will still work as before when `requestTimeout` is explicitly set to a value.


JoshMock:
I need to working on a change to the `body` parameter removal for 9.0.0-alpha.2, and then can merge this change for alpha.3.

JoshMock:
asciidoc failure is expected and being tracked elsewhere.

PR #2667, opened 2025-03-21
SPDX license format

JoshMock:
Switch to SPDX license format for all non-codegen files, and run a test to verify all code files have a license header.


JoshMock:
SPDX header check test is an expected failure. It should succeed on `main` on the next codegen run.

PR #2673, opened 2025-03-21
Drop broken support for roles

JoshMock:
Also update docs to reflect its true default value and how to use it properly. See <https://github.com/elastic/elastic-transport-js/issues/231>


JoshMock:
Not going to merge this yet. Latest discussion on [the community forum](https://discuss.elastic.co/t/node-js-client-uses-master-only-nodes-by-default/376162) makes it clear that reintroducing the behavior is preferred over dropping it.

JoshMock:
Killing this in favor of bringing this feature back.

PR #2697, opened 2025-04-04
Update helpers to use new multisearch types

JoshMock:
- **Update helpers to use correct multisearch types**
- **Stop copying project files to Dockerfile**


JoshMock:
CI failures are a chicken-and-egg problem that will be fixed on the next successful code generation, which is currently failing without the fix this PR makes. Will immediately rerun codegen after merging this.

PR #2717, opened 2025-04-07
Restore title from release note setup PR

marciw:
This PR reverts the navigation title change in https://github.com/elastic/elasticsearch-js/pull/2713, to what @KOTungseth did in https://github.com/elastic/elasticsearch-js/pull/2665 -- so that it looks like the other client docs.

right now it's a mismatch:
Screenshot 2025-04-07 at 6 03 13 PM


JoshMock (approved):
Thanks for the cleanup @marciw! Didn't realize the titles were being used elsewhere to roll up multiple clients' docs.

PR #2738, opened 2025-04-14
Update ESQL helper types

JoshMock:
The ESQL helper needs to reflect the types added to the spec in <https://github.com/elastic/elasticsearch-specification/pull/4189>


JoshMock:
Another chicken-and-egg typing issue: merging this will allow codegen to succeed, which will introduce the types that it is expecting to find.

PR #2756, opened 2025-04-16
Add test to verify default node filter function

JoshMock:
Turns out https://github.com/elastic/elastic-transport-js/issues/231 doesn't require any actual fix in the client! Adding a test and updating docs anyway.


JoshMock:
Tests will fail until the next transport patch release goes out.

Issue #3172, opened 2026-02-09
ESM dual package export (9cf1a34) breaks bundling


artemtam:
### 🐛 Bug report

`createRequire` in ESM build breaks bundling (9cf1a34). I tested with Bun and Rolldown (with node). Both can not properly resolve imported files which result in "Cannot find module" error.

```
node:internal/modules/cjs/loader:1453
  const err = new Error(message);
              ^

Error: Cannot find module '../package.json'
```

### To reproduce

Create a `repro.js` with a single import:

```js
import { Client } from '@elastic/elasticsearch';
```

Compile it with Bun or Rolldown:

```shell
bun build --target bun ./repro.js --outfile ./bundled.js

# or

npx rolldown --platform node ./repro.js > ./bundled.js
```

Run it:

```
bun ./bundled.js

# or

node ./bundled.js
```

Observe "Cannot find module" error.

### Expected behavior

It should work: bundlers should be able to analyze & resolve imported modules.

### Node.js version

v25.6.0

### @elastic/elasticsearch version

9.3.0

### Operating system

MacOS Tahoe 26.2 (ARM64)

margaretjgu:
Hi @artemtam,
Thanks for the detailed report and repro steps!
This has been fixed in `@elastic/elasticsearch@9.3.1` (and `@elastic/transport@9.3.3`). Please let us know if it doesn't resolve your problem.
If you're using a lockfile that still resolves @elastic/transport to 9.3.2, you may need to update it explicitly to pick up 9.3.3. Running npm update @elastic/transport or reinstalling should do it. Thanks!

artemtam:
Hi @margaretjgu! I can confirm, it is fixed in 9.3.1. 
Thanks a lot for being that quick!

PR #3175, opened 2026-02-09
fix bundle issue

margaretjgu:
Fixes https://github.com/elastic/elasticsearch-js/issues/3172
When apps use ESM (`import { Client } from '@elastic/elasticsearch'`) are bundled (e.g. with Bun, Rolldown, or esbuild) and then run...they can hit:
- `Cannot find module '../package.json'`, or
- `TypeError: The argument 'filename' must be a file URL... Received undefined`

**Why it failed:** 
The esm build of the client needs the package version (and transport version) for the `User-Agent` and `x-elastic-client-meta`. In ESM you can’t use `require()`, so the build was patched to use node’s `createRequire`
That works when the esm file is run as a real file on disk: `import.meta.url` is that file’s URL, so `createRequire(import.meta.url)` resolves `../package.json` relative to the installed package. But when the same ESM code is bundled into a single file (e.g. bundled.js in the app):
- The code that runs is inside the bundle, not in `node_modules/@elastic/elasticsearch/esm/`.
- Either `import.meta.url` is the bundle’s url (so ../package.json points to a path that doesn’t exist), or the bundler doesn’t provide a valid `import.meta.url` so its undefined.
- So `require('../package.json')` either fails with `“Cannot find module”` or `createRequire` fails because the filename is invalid.

TLDR -  any use of `createRequire(import.meta.url)` AND a path relative to the original package breaks when that code is bundled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "zx": "8.8.5"
   },
   "dependencies": {
-    "@elastic/transport": "^9.3.2",
+    "@elastic/transport": "^9.3.3",
     "apache-arrow": "18.x - 21.x",
     "tslib": "^2.4.0"
   }


prodsecmachine:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scanner |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clients-team/pr-checks/acb9df36-b875-4ffe-b32f-0bc7f953f82c?refs=b9cec2c8e80b2643f89b52e3f912db766e9a2a12&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clients-team/pr-checks/acb9df36-b875-4ffe-b32f-0bc7f953f82c/license?refs=b9cec2c8e80b2643f89b52e3f912db766e9a2a12&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

margaretjgu:
bundle ci fails because transport has the same issue. will open up a pr in transport to fix the same issue + do a release...client bundle check will then self resolve

PR #3380, opened 2026-07-23
feat: expose maxPathLength client option

gsoldevila:
## Summary
- Adds `ClientOptions.maxPathLength` and forwards it to `@elastic/transport` `TransportOptions.maxPathLength`.
- When set, the transport rejects requests whose `path` + `?querystring` byte length exceeds the limit **before** sending (avoids EPIPE / 414 from Elasticsearch's ~4kb `http.max_initial_line_length`).
- Does **not** include scheme or hostname (origin-form request-target only).
- Disabled by default (`null`) so consumers can choose the budget — Kibana can set e.g. `3500` in `configureClient` / `parseClientOptions` once this and the transport change are released and bumped.

Depends on: https://github.com/elastic/elastic-transport-js/pull/400

## Test plan
- [x] Unit test `Passes maxPathLength to Transport`
- [x] Full unit suite locally (640 pass, Node 22)
- [ ] After transport publish + bump: `new Client({ maxPathLength: 3500, ... })` rejects oversized `indices.resolveIndex` paths with `ConfigurationError`

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "zx": "8.8.5"
   },
   "dependencies": {
-    "@elastic/transport": "^9.3.5",
+    "@elastic/transport": "^9.4.0",
     "tslib": "^2.4.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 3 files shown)


margaretjgu:
transport has been released with v9.4 
https://github.com/elastic/elastic-transport-js/releases/tag/v9.4.0

PR #3394, opened 2026-07-30
fix(ci): create GitHub release after npm publish

margaretjgu:
The publish workflow was missing the GitHub release creation step. Adds it after npm publish, using the ephemeral vault token. Also bumps `contents` permission from `read` to `write` which is required to create releases.


JoshMock (commented):
The GH release creation step was removed now that we have an upstream "create release" workflow in the team repo that should trigger this downstream `npm publish`. If it does need to be reinstated, can we add a condition that it only runs when this workflow was triggered by a `workflow_dispatch`?

PR #3401, opened 2026-08-03
fix(ci): add --allow-empty-coverage to integration test tap run

margaretjgu:
`--disable-coverage` prevents c8 from collecting coverage data but doesn't gate tap's built-in report teardown. After all tests complete, `execute-test-suite.js` runs `report()` as a teardown, which checks `.tap/coverage` and exits 1 if it's empty - even when coverage was intentionally disabled. Adding `--allow-empty-coverage` suppresses that exit.

### DIFF
--- a/.buildkite/run-client.sh
+++ b/.buildkite/run-client.sh
@@ -35,7 +35,7 @@ docker run \
   --name elasticsearch-js \
   --rm \
   elastic/elasticsearch-js \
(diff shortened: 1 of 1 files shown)


margaretjgu:
Closing - this just adds --allow-empty-coverage which is another bypass. Will implement proper coverage with thresholds instead.

PR #3402, opened 2026-08-03
fix(ci): enable coverage with thresholds for integration tests

margaretjgu:
The integration tests were using `--disable-coverage` which prevented coverage collection but caused tap to exit 1 anyway - tap's built-in report teardown always runs after `tap run` and exits 1 when `.tap/coverage` is empty, regardless of `--disable-coverage`.

This replaces the bypass with actual coverage:
- Removes `--disable-coverage`
- Adds `--coverage-exclude` to exclude test files from measurement (we care about library code coverage, not test file coverage)
- Adds `--allow-incomplete-coverage` since integration tests won't load every source file
- Sets starting thresholds: lines=50, branches=30, functions=60, statements=50

The thresholds are conservative starting points to be tightened once we see actual numbers from the first green CI run. The unit test command (`npm test`) tracks the same library code at 96/90/96/91 via c8; these are lower because integration tests cover happy paths but not error handling or retry logic.

The OOM from a previous attempt came from wrapping tap with c8. This approach uses tap's native V8 coverage (each test file spawns its own process) which avoids that issue.


JoshMock (dismissed):
This LGTM btw! Been sitting in draft for 2 weeks so I kept overlooking it.

PR #3410, opened 2026-08-06
fix: prevent unhandled rejection in SniffingTransport error handling

### LINKED ISSUE #3407: SniffingTransport: catch handler assumes err.meta exists, can cause unhandled rejection and swallow the sniff diagnostic event
## Summary

`SniffingTransport.sniff()` can produce an unhandled promise rejection (and silently swallow the `sniff` diagnostic event) when the error caught in its `.catch` handler is not a transport error carrying a `meta` property.

## Details

https://github.com/elastic/elasticsearch-js/blob/8ae1fa35d9d8904aa9507ff048c1b3194fa2b0da/src/sniffingTransport.ts#L19-L35

Two problems compound here:

1. The `isObject` guard passes for `null`, because `typeof null === 'object'`:

   https://github.com/elastic/elasticsearch-js/blob/8ae1fa35d9d8904aa9507ff048c1b3194fa2b0da/src/sniffingTransport.ts#L38-L40

   So a `null` body slips past the `assert` and `nodesToHost(result.body.nodes, ...)` throws a `TypeError` inside the `.then` handler.

2. The `.catch` handler assumes every error has a `meta` object:

   ```ts
   .catch(err => {
     this.isSniffing = false
     err.meta.sniff = { hosts: [], reason: opts.reason }  // <-- throws if err is a TypeError/AssertionError
     this.diagnostic.emit('sniff', err, null)
   })
   ```

   Transport errors (`ResponseError`, `ConnectionError`, etc.) do have `meta`, but a `TypeError` from (1), or the `AssertionError` from the `assert(isObject(...))` line, does not. `err.meta.sniff = ...` then throws a *second* error inside the `.catch`, which nothing handles.

## Impact

`sniff()` is fire-and-forget (nothing awaits the promise chain), so the secondary throw becomes an unhandled promise rejection — which terminates the process by default on modern Node.js — and the `sniff` diagnostic event is never emitted. Likelihood is low in normal operation (requires a malformed/unexpected sniff response body), but the failure mode is a crash in a background code path that gives no indication of the root cause.

## Suggested fix

- Make the guard reject `null`: `typeof obj === 'object' && obj !== null`
- Make the `.catch` defensive about non-transport errors, e.g. only attach `sniff` metadata when `err.meta` exists (or wrap the error), so `diagnostic.emit('sniff', err, null)` is always reached.

Found while auditing the client code; happy to open a PR if the proposed direction sounds right.

ch-bas:
Closes #3407

## What this does

`SniffingTransport.sniff()` could turn a background sniff failure into an unhandled promise rejection, because its `.catch` handler assumed every error carries a transport `meta` object:

```ts
.catch(err => {
  this.isSniffing = false
  err.meta.sniff = { hosts: [], reason: opts.reason }  // throws if err has no meta
  this.diagnostic.emit('sniff', err, null)             // never reached
})
```

Transport errors do have `meta`, but errors thrown inside the `.then` handler do not — e.g. the `AssertionError` from `assert(isObject(result.body))`, or a `TypeError` from `nodesToHost` when the body has no `nodes`. Since `sniff()` is fire-and-forget, the secondary throw became an unhandled rejection (fatal by default on modern Node.js) and the `sniff` diagnostic event was silently swallowed.

Two changes:

1. The `.catch` only attaches sniff metadata when `err.meta` exists, so `diagnostic.emit('sniff', err, null)` is always reached.
2. `isObject` now rejects `null` (`typeof null === 'object'`), so a JSON `null` sniff body fails the assertion instead of slipping through to `result.body.nodes`.

## Tests

Two new unit tests in `test/unit/sniffingTransport.test.ts` using `buildMockConnection`:

- a sniff response with a JSON `null` body
- a sniff response without a `nodes` property

Both assert the `sniff` diagnostic event fires with an error and `isSniffing` is reset. Verified both tests fail against the previous implementation (the unhandled rejection kills the test process) and pass with this change. Full `npm test` passes cleanly with coverage thresholds met.

### DIFF
--- a/src/sniffingTransport.ts
+++ b/src/sniffingTransport.ts
@@ -29,12 +29,16 @@ export default class SniffingTransport extends Transport {
       })
       .catch(err => {
         this.isSniffing = false
-        err.meta.sniff = { hosts: [], reason: opts.reason }
+        // non-transport errors, like the AssertionError thrown on an
+        // unexpected body, don't have a meta object to attach sniff data to
+        if (isObject(err?.meta)) {
+          err.meta.sniff = { hosts: [], reason: opts.reason }
+        }
         this.diagnostic.emit('sniff', err, null)
       })
   }
 }
 
 function isObject (obj: any): obj is Record<string, any> {
-  return typeof obj === 'object'
+  return typeof obj === 'object' && obj !== null
 }
--- a/test/unit/sniffingTransport.test.ts
+++ b/test/unit/sniffingTransport.test.ts
@@ -60,5 +60,51 @@ test('SniffingTransport', t => {
     transport.sniff({ reason: 'test' })
   })
 
+  t.test('sniff - null body emits sniff event with error', t => {
+    t.plan(2)
+
+    const MockConnection = connection.buildMockConnection({
+      onRequest (_params) {
+        return { body: null }
+      }
+    })
+
+    const client = new Client({
+      node: 'http://localhost:9200',
+      Connection: MockConnection
+    })
+
+    const transport = client.transport as SniffingTransport
+    transport.diagnostic.once('sniff', (err, result) => {
+      t.ok(err != null, 'the sniff event should carry the error')
+      t.equal(transport.isSniffing, false)
+    })
+
+    transport.sniff({ reason: 'test' })
+  })
+
+  t.test('sniff - body without nodes emits sniff event with error', t => {
+    t.plan(2)
+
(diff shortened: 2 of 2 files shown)


margaretjgu:
left a note on #3407 here: https://github.com/elastic/elasticsearch-js/issues/3407#issuecomment-5207623940. closing that one as not planned given it's an edge case, so closing this pr for now too. feel free to take a gander at our `good first issue` label if you're looking for something else to pick up.

PR #3420, opened 2026-08-24
fix: codegen

margaretjgu:
fix codegen


margaretjgu:
closing, generator fix first

PR #3424, opened 2026-08-26
docs: add claude test skill

margaretjgu:
Test stage is still 0. Adds `.claude/skills/test.md` so the collector can score Experimenting (level 1).


JoshMock (approved):
LGTM

PR #3434, opened 2026-09-02
Make SecurityGlobalPrivilege application attribute optional

mattkime:
The application attribute of SecurityGlobalPrivilege should be optional, here's an explain of it being omitted - https://www.elastic.co/docs/reference/query-languages/esql/esql-data-federation-security#privileges


JoshMock (approved):
LGTM. Made the same change to the spec to ensure it's permanent. The file you edited is code-generated so the next codegen run would have reverted your fix. https://github.com/elastic/elasticsearch-specification/pull/6625

PR #3435, opened 2026-09-04
fix: move Apache Arrow optional type defs out of default import tree

### LINKED ISSUE #3419: esm/helpers.d.ts imports 'apache-arrow/Arrow.node.js' which can't be resolved
### 🐛 Bug report

using the ES|QL Arrow helpers from an ESM TypeScript project doesnt compile. The shipped `esm/helpers.d.ts` imports the Arrow types from `apache-arrow/Arrow.node.js` and that path can't be resolved. `lib/helpers.d.ts` uses `apache-arrow/Arrow.node` without the extension and works fine, so this only affects ESM consumers .

With `skipLibCheck` off (TypeScript's default) tsc fails with TS2307. With `skipLibCheck` on the error disappears, but `Table` ends up as `any`, so a typo on the table object compiles clean. The same file in a CJS project catches it.

I had a look at why. `tsc-esm-fix` adds the `.js` to both `esm/helpers.js` and `esm/helpers.d.ts`, and `scripts/fix-esm-require.js` only puts it back for the runtime filebecause its regex looks for `require('apache-arrow/Arrow.node.js')`. apache-arrow maps `./*` to `./*.d.ts`, so `Arrow.node.js` would need an `Arrow.node.js.d.ts`, which doesn't exist. Only `Arrow.node.d.ts` does.

Adding the same replacement for `esm/helpers.d.ts` in that script fixes both cases for me
locally. I will send a PR if thats the direction you want.

### To reproduce

An ESM TypeScript project with apache-arrow installed:

```bash
mkdir es-arrow-esm && cd es-arrow-esm
npm init -y
npm pkg set type=module
npm i @elastic/elasticsearch@9.5.0 apache-arrow@21 typescript@5.9.3 @types/node@22
```

```ts
// fixture.ts

export async function go () {
  const table = await client.helpers.esql({ query: 'FROM x' }).toArrowTable()
  return table.numRows
}
```

```bash
npx tsc fixture.ts --module nodenext --moduleResolution nodenext --target es2022 --strict --noEmit
```

```
node_modules/@elastic/elasticsearch/esm/helpers.d.ts(3,67): error TS2307: Cannot find module 'apache-arrow/Arrow.node.js' or its corresponding type declarations.
```

The two declaration files disagree, and the runtime ESM file already has the right specifier:

```
$ grep -n apache-arrow node_modules/@elastic/elasticsearch/lib/helpers.d.ts
3:import type { Table, TypeMap, AsyncRecordBatchStreamReader } from 'apache-arrow/Arrow.node';

$ grep -n apache-arrow node_modules/@elastic/elasticsearch/esm/helpers.d.ts
3:import type { Table, TypeMap, AsyncRecordBatchStreamReader } from 'apache-arrow/Arrow.node.js';

$ grep -n apache-arrow node_modules/@elastic/elasticsearch/esm/helpers.js
14:        return require('apache-arrow/Arrow.node');
```

Turning on `skipLibCheck` hides the error instead of fixing it. With `thisPropertyDoesNotExist`, which isn't a real property on Arrow's `Table`:

```ts
// fixture2.ts

export async function go () {
  const table = await client.helpers.esql({ query: 'FROM x' }).toArrowTable()
  return table.thisPropertyDoesNotExist
}
```

```bash
npx tsc fixture2.ts --module nodenext --moduleResolution nodenext --target es2022 --strict --noEmit --skipLibCheck
```

```
ESM project ("type": "module"):  exit 0, no error
CJS project (no "type" field):   fixture2.ts(7,16): error TS2339: Property 'thisPropertyDoesNotExist' does not exist on type 'Table<TypeMap>'.
```

So an ESM project either won't compile, or compiles with no Arrow typing at all.

No Elasticsearch instance needed. Same on main (dcc98b2e), and the specifier is the same in 9.3.0 through 9.5.0.

### Expected behavior

`esm/helpers.d.ts` should use the same specifier as `lib/helpers.d.ts` (`apache-arrow/Arrow.node`), so ESM consumers get the Arrow types that CJS consumers already get.

### Node.js version

22.22.2

### @elastic/elasticsearch version

9.5.0

### Operating system

macOS 26.6

### Any other relevant environment information

TypeScript 5.9.3, apache-arrow 21.1.0, @elastic/transport 9.4.0. Consumer project is "type": "module" with moduleResolution nodenext.

JoshMock:
Fixes https://github.com/elastic/elasticsearch-js/issues/3419

### DIFF
--- a/docs/reference/client-helpers.md
+++ b/docs/reference/client-helpers.md
@@ -634,6 +634,27 @@ const result = await client.helpers
   .toRecords<EventLog>()
 ```
 
+#### Loading Arrow type definitions [_arrow-types]
+
+`apache-arrow` is an optional peer dependency, so its type definitions are not wired into the client by default. This keeps projects that never use the Arrow helpers from having to install `apache-arrow` to type-check. Until the definitions are loaded, `toArrowTable` returns `unknown` and `toArrowReader` returns `unknown`; both still work at runtime.
+
+To get precise types (`Table` and `AsyncRecordBatchStreamReader`), install `apache-arrow` and add a single side-effect import once, anywhere in your project:
+
+```ts
+// Loads the apache-arrow type definitions for the ES|QL Arrow helpers.
+// A side-effect import: no bindings are needed, and it is erased at runtime.
+import '@elastic/elasticsearch/helpers-arrow'
+
+import { Client } from '@elastic/elasticsearch'
+
+const client = new Client({ node: 'http://localhost:9200' })
+
+// `table` is now typed as `Table`, and `reader` as `AsyncRecordBatchStreamReader`
+const table = await client.helpers
+  .esql({ query: 'FROM sample_data' })
+  .toArrowTable()
+```
+
 #### `toArrowReader` [_toarrowreader]
 
 Added in `v8.16.0`
--- a/docs/reference/client-helpers.md
+++ b/docs/reference/client-helpers.md
@@ -642,7 +642,8 @@ To get precise types (`Table` and `AsyncRecordBatchStreamReader`), install `apac
 
 ```ts
 // Loads the apache-arrow type definitions for the ES|QL Arrow helpers.
-// A side-effect import: no bindings are needed, and it is erased at runtime.
+// This is a type-only augmentation: at runtime it just loads an empty module.
+// It imports no bindings, so nothing else needs to change.
 import '@elastic/elasticsearch/helpers-arrow'
 
 import { Client } from '@elastic/elasticsearch'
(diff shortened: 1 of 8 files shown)


margaretjgu (dismissed):

seialkali:
I ran this PR through an LLM and it flagged something that might be worth a look. I don't know this codebase well enough to judge, but I wanted to pass it along just in case it's useful:

> The concern is projects on `moduleResolution: node10`. That mode predates `exports` maps and ignores them, and since the client keeps its declarations in `lib/` and `esm/` with only `index.d.ts` at the package root, subpath imports aren't reachable from it. So `import '@elastic/elasticsearch/helpers-arrow'`, which the breaking changes note gives as the way to restore the Arrow types, may not resolve for those projects.
> 
> ```
> $ npx tsc --noEmit --module commonjs --moduleResolution node10 --skipLibCheck probe.ts
> probe.ts(1,26): error TS2307: Cannot find module '@elastic/elasticsearch/helpers' or its corresponding type declarations.
>   There are types at '.../esm/helpers.d.ts', but this result could not be resolved under your current
>   'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.
> ```
> That same setup type-checks `toArrowTable().numRows` fine today, so those projects would end up with `unknown` and no documented way to opt back in. Might just be worth mentioning the node16/nodenext/bundler requirement in the breaking changes note.

